PR #337, opened 2014-12-29
Add special handling for pistons in global region; Fixes WORLDGUARD-3195

PseudoKnight:
I'm not sure if this is how you want to solve this problem, but I thought I'd give this issue a nudge and offer a possible solution. This approach is mostly to avoid unintended side-effects and work as seemingly intended. Getting source to contain global would also work, but it's less obvious where I should change that.


sk89q:
Honestly, this is what I was going to do until I remembered to check out your PR.

It does mean a special case exception in the code but I think it's reasonable.

PR #338, opened 2015-01-09
Added getApplicableRegions(Chunk chunk);

jjssman:
Added the method getApplicableRegions(Chunk chunk) to the API: http://youtrack.sk89q.com/issue/WORLDGUARD-3292


sk89q:
I added BukkitUtil.toRegion(Chunk) instead.

PR #339, opened 2015-01-14
Added animal damage flag

DerFlash:
Due to the latest 6.0 changes we had the problem that users could no longer damage entities, which they could before. We saw that this was checked by the interact flag, but enabling this one globally would cause other bigger problems.
Basically we're running with a global build deny, having cities and inner user zones with special build flags. Thus we just needed to be able to allow entity damage again. So I thought adding a special flag should be the right way!


wizjany:
fixed in cf1716ff372003a8ba658fd2451839b9d23c1894

PR #340, opened 2015-02-16
Make RegionPrintoutBuilder use the FlagRegistry

ZephireNZ:
Currently, when running ```/region info``` or any other command that lists a region flags, it will only show default flags. This simple PR changes that, by instead making it iterate over the FlagRegistry.

I'm not certain on what the standard is for getting the WorldGuardPlugin instance, please correct me if I'm wrong.

### DIFF
--- a/src/main/java/com/sk89q/worldguard/bukkit/commands/region/RegionPrintoutBuilder.java
+++ b/src/main/java/com/sk89q/worldguard/bukkit/commands/region/RegionPrintoutBuilder.java
@@ -21,10 +21,12 @@ package com.sk89q.worldguard.bukkit.commands.region;
 
 import com.sk89q.squirrelid.cache.ProfileCache;
 import com.sk89q.worldedit.BlockVector;
+import com.sk89q.worldguard.bukkit.WorldGuardPlugin;
 import com.sk89q.worldguard.domains.DefaultDomain;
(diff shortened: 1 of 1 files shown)


ZephireNZ:
Note: build is failing because the project uses APIs only accessible in JDK 8. Switching to opendk8 (as I had to do on my own machine) fixes the build issues.

wizjany:
We still target java6 though.

ZephireNZ:
Indeed, however previous builds (which run just fine) are also erroring on Travis as well: https://travis-ci.org/sk89q/WorldGuard/branches

The java 8 compiled versions still seem to work just fine in java 7, despite referencing java 8 libraries.

wizjany:
I'm confused. master currently builds fine on java6, right?

ZephireNZ:
Yep. You yourself commited 9998a6c to fix compatibility issues, however the feature/custom-flags branch is based on older commit, one that does not have the fix, and hence does not compile on anything less than java 8.

wizjany:
Ah ok. It's just builds of this branch.

ZephireNZ:
Once the custom-flags branch is either merged or rebased, it should be fine.

wizjany:
merged to feature/flagcontext branch

PR #341, opened 2015-04-19
Add support for Armor stands in protected regions

MrStickmanPro1:
Add support for armor stands in regions with the "mob-spawning: deny" flag by cancelling the onCreatureSpawn method and therefor not preventing them to be spawned.


wizjany:
look down 20 lines

MrStickmanPro1:
Sorry, did not see that... I still might have another thing about placing armor stands tho - will look into that later.

PR #343, opened 2015-05-31
Remove action in blacklist

johandrex:
Hello, I wonder if you could add an action in the blacklist "remove". Which should remove the item the user is trying to take from a chest, enderchest, inventory or pickup. That'd be awesome.


wizjany:
feature requests go on the [issue tracker](http://http://youtrack.sk89q.com)

PR #344, opened 2015-06-05
FEATURE OR FLAG ADJUSTMENT URGENT

RAZERMC:
Please make the PVP flag respect the outside and inside of the region, meaning i create a region then flag it as pvp allow. then restrict players to harm another player when inside the region. Right now players outside the region can do damage to the players inside the pvp region. suggestion put a check if the source attacking player is inside the same region if true then allow the event if false cancel the event and say "your are not allowed to pvp here"

### DIFF
--- a/src/main/java/com/sk89q/worldguard/bukkit/commands/region/RegionCommands.java
+++ b/src/main/java/com/sk89q/worldguard/bukkit/commands/region/RegionCommands.java
@@ -44,11 +44,7 @@ import com.sk89q.worldguard.bukkit.commands.task.RegionRemover;
 import com.sk89q.worldguard.bukkit.permission.RegionPermissionModel;
 import com.sk89q.worldguard.bukkit.util.logging.LoggerToChatHandler;
 import com.sk89q.worldguard.protection.ApplicableRegionSet;
-import com.sk89q.worldguard.protection.flags.DefaultFlag;
-import com.sk89q.worldguard.protection.flags.Flag;
-import com.sk89q.worldguard.protection.flags.InvalidFlagFormat;
-import com.sk89q.worldguard.protection.flags.RegionGroup;
-import com.sk89q.worldguard.protection.flags.RegionGroupFlag;
+import com.sk89q.worldguard.protection.flags.*;
 import com.sk89q.worldguard.protection.managers.RegionManager;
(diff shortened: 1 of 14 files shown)


DarkArc:
Hi RAZERMC, thanks for voicing your concern, but this isn't the place to do it. We have a dedicated issue tracker located [here](http://youtrack.sk89q.com).

It should be checking both player's regions, and seeing if either region does not allow PvP. If one player's region set doesn't allow PvP, then it will block the combat. If that's not what's happening, feel free to post a bug report -- though these things generally are some kind of user error. Additionally, if [this issue](http://youtrack.sk89q.com/issue/WORLDGUARD-3494) is yours, it might be a good idea to add additional detail to your report, as in it's current state another developer -- I don't work on Bukkit plugins anymore -- will likely mark it as can't reproduce.

On the other hand, if that's not what you want to happen, you can feel free to file a feature/enhancement request. There's no pro misses we'll implement it, but if it's a good idea, and it doesn't require a large amount of effort, it has a fair chance to get developed. As is the case with open source software, you can of course create your own fork, and add the feature you need as well. 

Also, just a bit of advice, in all the open source projects I've been active in, I've never once seen writing something in all caps be beneficial to the author's cause. It tends to make you look a bit self entitled, and since we don't make money off of you, and merely share our code for free, it's important to note that in no way are we obligated to help you, although many of us still will. It's far better to share your idea in a calm and collected manor, or implement the feature yourself, and send us a pull request! :)

wizjany:
yea, he came on irc and i directed him there

DarkArc:
Ah alright, you beat me by half a second closing this lol.

RAZERMC:
im greatly sorry for the false report after intensive testing it was pvpmanger that was causing the bug my bad sorry for that well more support guys!

RAZERMC:
already informed the pvpmanager author about the issue and have requested worldguard region checks feature they only support faction. again i apologize for the false report.

TomyLobo:
Can we delete this PR so someone doesn't accidentally click that "Delete branch" button and delete the branch @RAZERMC randomly pullrequested?

DarkArc:
Or you know, we an just leave it in the shadows, and it'll go away.

PR #346, opened 2015-07-07
Use UUID's for hostname based security

Bestle:
Pretty sure this is the only place a change is required to support UUID's.


zml2008:
This will break existing configurations -- you should probably hook this up to look up both by uuid and name, giving uuid priority, or doing whatever it is WG does for region specifications to handle both (and convert if necessary?).

TomyLobo:
Better yet, instead of copying code over, extract a method and reuse it.

wizjany:
as noted, this breaks existing configs.

PR #348, opened 2015-12-04
Versions/spigot1.8 5.x

JugadorNinja:
we cant break some block like slime blocks, primarine blocks on 1.8.8 server using WorldGuard 6.1


DarkArc:
This is not the place to ask for help/report a bug.

PR #349, opened 2016-01-05
Reset player saturation and exhaustion for the /heal command.

kosakriszi:
This is a pretty simple PR, it modifies the /heal command to reset the target's saturation and exhaustion levels, so you don't immediately start losing hunger after a /heal.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 jdk:
-  - oraclejdk7
-  - openjdk7
-  - openjdk6
+  - oraclejdk8
 notifications:
   email: false
+before_install: chmod +x gradlew
+script: ./gradlew build -S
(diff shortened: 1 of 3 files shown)


kosakriszi:
Looks like there's some wonky permission denied errors in the Travis system.

TomyLobo:
Maybe these should be flags? what do the others think?
Maybe a flag for health as well that defaults to true only if none of the others are set?
"/heal -hfe" or something

btw, there shouldn't be any merge commits in your PR. please rebase it.

DarkArc:
I'm fine with just merging this as is. This wasn't updated because the command was deprecated in favor of CommandBook's version, which does pretty much exactly the same thing. Updating saturation is a reasonable default behavior. 
		
        
		
		 ---- On Wed, 06 Jan 2016 14:03:13 -0800  notifications@github.com  wrote ----Maybe these should be flags? what do the others think?
Maybe a flag for health as well that defaults to true only if none of the others are set?
"/heal -hfe" or something
btw, there shouldn't be any merge commits in your PR. please rebase it.

TomyLobo:
oh right, i forgot. no objections to merging then

kosakriszi:
I'd love to rebase, but I have no idea how to do it properly, and Google honestly isn't helping me.

DarkArc:
@kosakriszi here, see if you're interested in learning to rebase, see if this helps you: http://haxing.ninja/tutorial/git/rebase/. I quickly wrote up a tutorial for git rebasing, particularly for situations like this, it could stand some improvement, but it should be useful none the less.

kosakriszi:
Thank you for the guide, I will remake this pull request, since I've sufficiently messed this up with my fooling around

PR #350, opened 2016-01-07
Reset player saturation and exhaustion for heal command

kosakriszi:
This is a remake of my old pull request (#349), now with 120% more squashed commits! Thanks again @DarkArc for your tutorial!


DarkArc:
Glad it helped, you could have force pushed (````git push -f````) to update/overwrite the old branch -- and thus the old PR -- though, just for future reference.

kosakriszi:
Noted, I had thought I messed it up beyond repair.

PR #347, opened 2015-08-02
Custom flags and context and stuff

wizjany:
Lets us do some more advanced stuff and lets other plugins register flags and session handlers.

PR #353, opened 2016-05-03
Added experimental custom flag support

Challenger2:
The custom flags work as follows:

1. WorldGuard's onEnable() is called.

2. World Guard looks for any plugin that implements CustomFlagProvider

3. World Guard invokes addCustomFlags() on each found plugin

4. World Guard passes in a CustomFlagBroker.

5. The plugin calls CustmFlagBroker.addFlag to add any flags it needs

A plugins addFlag method will be called after onLoad(), but may be called
before onEnable().

If a plugin that adds custom flags fails to load or is removed, World Guard
will remove all of the custom flags from all regions.
(Perhaps this can be changed in the future)

The above procedure fixes certain problems with WGCustomFlags, where WGCustomFlags tries to update the WorldGuard flags at the same time WorldGuard is loading asynchronously. WGCustomFlags cannot inject flags after WorldGuard load but before WorldGuard loads in its files. This was the primary motivation to adding a CustomFlagProvider and allowing WorldGuard itself to solicit custom flags from plugins.

### DIFF
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
@@ -40,6 +40,9 @@ import com.sk89q.worldguard.bukkit.event.player.ProcessPlayerEvent;
 import com.sk89q.worldguard.bukkit.listener.*;
 import com.sk89q.worldguard.bukkit.util.Events;
 import com.sk89q.worldguard.protection.GlobalRegionManager;
+import com.sk89q.worldguard.protection.flags.CustomFlagBroker;
+import com.sk89q.worldguard.protection.flags.CustomFlagBrokerImpl;
+import com.sk89q.worldguard.protection.flags.CustomFlagProvider;
 import com.sk89q.worldguard.protection.flags.Flag;
 import com.sk89q.worldguard.protection.managers.RegionManager;
 import com.sk89q.worldguard.protection.managers.storage.StorageException;
@@ -129,6 +132,19 @@ public class WorldGuardPlugin extends JavaPlugin {
         configureLogger();
 
         getDataFolder().mkdirs(); // Need to create the plugins/WorldGuard folder
+        
+        // Load in custom flags
+        for (Plugin plugin : this.getServer().getPluginManager().getPlugins()) {
+            CustomFlagBroker broker = new CustomFlagBrokerImpl();
+            if (plugin instanceof CustomFlagProvider) {
+                CustomFlagProvider provider = (CustomFlagProvider)plugin;
+                try {
+                    provider.addCustomFlags(broker);
+                } catch (Exception e) {
+                    // Ignore
+                }
+            }
+        }
 
         executorService = MoreExecutors.listeningDecorator(EvenMoreExecutors.newBoundedCachedThreadPool(0, 1, 20));
 
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
@@ -131,20 +131,20 @@ public class WorldGuardPlugin extends JavaPlugin {
     public void onEnable() {
         configureLogger();
 
-        getDataFolder().mkdirs(); // Need to create the plugins/WorldGuard folder
-        
         // Load in custom flags
+        CustomFlagBroker broker = new CustomFlagBrokerImpl();
         for (Plugin plugin : this.getServer().getPluginManager().getPlugins()) {
-            CustomFlagBroker broker = new CustomFlagBrokerImpl();
             if (plugin instanceof CustomFlagProvider) {
                 CustomFlagProvider provider = (CustomFlagProvider)plugin;
                 try {
                     provider.addCustomFlags(broker);
                 } catch (Exception e) {
-                    // Ignore
+                    log.log(Level.WARNING, "Unhandled exception from plugin \"" + plugin + "\"", e);
                 }
             }
         }
+
+        getDataFolder().mkdirs(); // Need to create the plugins/WorldGuard folder
 
         executorService = MoreExecutors.listeningDecorator(EvenMoreExecutors.newBoundedCachedThreadPool(0, 1, 20));
 
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/DefaultFlag.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/DefaultFlag.java
@@ -19,16 +19,10 @@
 
 package com.sk89q.worldguard.protection.flags;
 
-import java.util.Arrays;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-
 import org.bukkit.ChatColor;
 import org.bukkit.GameMode;
 import org.bukkit.WeatherType;
 import org.bukkit.entity.EntityType;
-
-import com.sk89q.worldguard.bukkit.commands.region.RegionCommands;
 
 /**
  * The flags that are used in WorldGuard.
@@ -162,10 +156,6 @@ public final class DefaultFlag {
             TELE_LOC, SPAWN_LOC, POTION_SPLASH, TIME_LOCK, WEATHER_LOCK,
             BLOCKED_CMDS, ALLOWED_CMDS, PRICE, BUYABLE, ENABLE_SHOP
     };
-    
-    
-    // Remember if we have unlocked the flag field with reflection
-    private static boolean addFlagsUnlocked = false;
 
     private DefaultFlag() {
     }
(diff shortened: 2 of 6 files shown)


worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/DefaultFlag.java
```diff
@@ -179,4 +189,36 @@ private DefaultFlag() {
...
+        // we have to be able to change it too :/
+        try {
+            java.lang.reflect.Field field = DefaultFlag.class.getField("flagsList");
+            if (!addFlagsUnlocked) {
+                addFlagsUnlocked = true;
+                java.lang.reflect.Field modifiersField = java.lang.reflect.Field.class.getDeclaredField("modifiers");
+                modifiersField.setAccessible(true);
+                modifiersField.setInt(field, field.getModifiers() & ~java.lang.reflect.Modifier.FINAL);
+            }
+            field.set(null, newList);
```
TomyLobo: Is there any particular reason why we set our own fields via reflection?

Challenger2: Yes. It might be a bit strange, but it seemed to be the best way to ensure full compatibility. If I made the field writeable, other plugins could change the flags at runtime and cause breakage. If I made it private and provided a gitFlags method, I would break about all plugins out there

The way it is now, no other plugin can change the flags. At the very least, they would have to use reflection to change it, such as WGCustomFlags. I had issues with WGCustomFlags not working, so I made this quick patch.

The addFlags method is package private, so it is not exposed to other plugins. The only time you can update the flags is when the server first loads via the callback.

Challenger2: An example of a plugin that uses this is here:
https://github.com/NerdNu/NerdFlags/blob/master/src/com/michaelelin/NerdFlags/NerdFlagsPlugin.java

TomyLobo: Then make the field private and add a getter.

worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/DefaultFlag.java
```diff
@@ -179,4 +189,36 @@ private DefaultFlag() {
 
         return null;
     }
+
+    /**
+     * Add a new custom flag to the flags list
+     *
+     * @param flag The flag to add
+     */
+    static void addFlag(Flag<?> flag) {
```
TomyLobo: Try to keep the default-visible things at a minimum.
This can just as well be invoked by using new CustomFlagBrokerImpl().addFlag(...), so hiding it by making it default-visible makes no sense anyway.

worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
```diff
@@ -129,6 +132,19 @@ public void onEnable() {
...
+        
+        // Load in custom flags
+        for (Plugin plugin : this.getServer().getPluginManager().getPlugins()) {
+            CustomFlagBroker broker = new CustomFlagBrokerImpl();
+            if (plugin instanceof CustomFlagProvider) {
+                CustomFlagProvider provider = (CustomFlagProvider)plugin;
+                try {
+                    provider.addCustomFlags(broker);
+                } catch (Exception e) {
+                    // Ignore
```
TomyLobo: You should at least print the exceptions.

Also, I could think of a safer way in the face of exceptions:
````diff
+                try {
+                    provider.clear(); // clear internal list
+                    provider.addCustomFlags(broker); // add flags to an internal list
+                    provider.commit(); // commit list to DefaultFlag.flagsList
+                } catch (Exception e) {
````

File: worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java [hunk 4]
```diff
@@ -129,6 +132,19 @@ public void onEnable() {
         configureLogger();
 
         getDataFolder().mkdirs(); // Need to create the plugins/WorldGuard folder
+        
+        // Load in custom flags
+        for (Plugin plugin : this.getServer().getPluginManager().getPlugins()) {
+            CustomFlagBroker broker = new CustomFlagBrokerImpl();
```
wizjany: you're creating a new object that won't get used for 99% of plugins.

worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/DefaultFlag.java
```diff
@@ -179,4 +189,36 @@ private DefaultFlag() {
...
+                addFlagsUnlocked = true;
+                java.lang.reflect.Field modifiersField = java.lang.reflect.Field.class.getDeclaredField("modifiers");
+                modifiersField.setAccessible(true);
+                modifiersField.setInt(field, field.getModifiers() & ~java.lang.reflect.Modifier.FINAL);
+            }
+            field.set(null, newList);
+        } catch (Exception e) {
+            Logger log = Logger.getLogger(DefaultFlag.class.getCanonicalName());
+            log.log(Level.WARNING, "Failed to add a custom flag", e);
+            throw new RuntimeException("Internel Error", e);
```
wizjany: Um, why? Seems a bit extreme.

wizjany:
I'm not really sure why you thought reflecting on our own fields is the correct way to go about this.

There's multiple issues with your code style and usage choices (a RuntimeException? really?) which TomyLobo and I have mostly pointed out.

Have you looked at all at https://github.com/sk89q/WorldGuard/pull/347/commits/45f316f9431d3f6c3587ab223d4d235a1ede702d ? Although it's still unfinished it covers a lot of things you don't, such as thread safety, registry locking, an actual API, error-checking, and fallbacks for missing plugins.

worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
TomyLobo: It's created once, in onEnable... after that, the GC eats it and it's gone.
Not a big deal.

worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/DefaultFlag.java
TomyLobo: Also, typo.
But my opinion is to get rid of this reflection anyway.

worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
wizjany: it's more a code style issue than a performance issue. he could move the line down by one and boom fixed.

TomyLobo: Oh I hadn't noticed it's done for every plugin. Yeah that is bad. I thought it was outside the loop...
But inside the if is probably better in conjunction with my suggestion in a comment a few lines below this one.

Challenger2:
Thanks for all your feed back.

I have adjusted the patch to take into account all your comments the best I could. I had a look at the [45f316f] patch, but it seems old and far more complicated.

The new patch only adds additional files and adds an optimized hook into onEnable(). The rest of the WG API is unchanged. It may be a bit of a hack, but it works and could be a good stop-gap until a better way is merged in.

wizjany:
Does "old and complicated" somehow imply worse?
Hacks like this PR belong in third party plugins, which is what WGCustomFlags has been doing the entire time. The only problem I've heard you actually present about the way WGCustomFlags is doing it (besides reflection being a hack) is something to do with thread issues, which you haven't really explained. Perhaps there's better ways to work around that issue without moving a block of reflection into worldguard.

Challenger2:
I'll try to explain the threading issue.

1. WorldGuard.OnEnable() is called
2. Starts Asynchronous tasks that begin to load region data
3. NerdFlags.OnEnable() is called
4. NerdFlags adds flags to WGCustomFlags
5. WGCustomFlags hacks flacks into WG.

The problem is, if the region data is loaded before or after step 5. The flags sometimes work on our servers, and sometimes they don't. The flags don't work at all on 1.9.

The patch ensures that the flags are hacked in before WorldGuard starts spawning off threads or even starts reading from disk. This way, WG has no idea new flags were added at all and everything works as expected.

Challenger2:
I have been considering how to fix the threading problem. We need to hack in the new flags after WorldGuard is loaded but before WorldGuard.onEnable() is called. *Maybe* this could be done in another plugin during onLoad()?

totemo:
I agree that using reflection to modify arrays in WorldGuard is not desirable.  Having a flag registry is the way to go.

WGCustomFlags is unreliable; it fails to register custom flags with monotonous regularity.  Our main concern is getting a reliable custom flag mechanism, and the way WGCustomFlags does it isn't.  It makes sense that WorldGuard should implement something. Putting the custom flags mechanism outside of WorldGuard basically guarantees that it will have to be some kind of hack, and unfortunately just isn't reliable.

If WorldGuard is able to retain knowledge of flags, even when the plugin that defines them fails to load, that would be fantastic, because we have been bitten by that - losing custom flags from a great many regions.  I look forward to whatever you guys can come up with.  Hopefully soon?

wizjany:
>If WorldGuard is able to retain knowledge of flags, even when the plugin that defines them fails to load, that would be fantastic, because we have been bitten by that - losing custom flags from a great many regions

sk already thought of that. https://github.com/sk89q/WorldGuard/commit/45f316f9431d3f6c3587ab223d4d235a1ede702d#diff-ac85358260e6e0038aa10870627f14f8R109

wizjany:
closing in favor of #347 as discussed, i see no reason to prefer this system over that one.
if there is something important you feel that implementation is missing, feel free to comment on it.

PR #356, opened 2016-11-09
Fix Materials.ARMOR_STAND being null in 1.8.

filoghost:
In 1.8 Materials.ARMOR_STAND is null because END_CRYSTAL throws NoSuchFieldError and in the catch block ARMOR_STAND is set to null. This causes side effects such as players being able to place armor stands in protected regions.

### DIFF
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/util/Materials.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/util/Materials.java
@@ -54,10 +54,7 @@ public final class Materials {
             // for backwards compatible access to material enum
             ARMOR_STAND = Material.ARMOR_STAND;
             END_CRYSTAL = Material.END_CRYSTAL;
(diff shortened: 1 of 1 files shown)


TomyLobo:
This changes nothing.
Uninitialized fields are null.

filoghost:
In 1.8, Materials.ARMOR_STAND is initialized correctly, then Materials.END_CRYSTAL fails (because the material was added in 1.9), and when catching the ignored exception it resets Materials.ARMOR_STAND to null.

TomyLobo:
I see. I assumed those were introduced in the same version.
So who wrote this nonsense? :D

wizjany:
I haven't played MC since 1.7.10, so I have no clue what version things were added in ¯\_(ツ)_/¯
I added the second part in https://github.com/sk89q/WorldGuard/commit/12fc1438c2deeb081669c7e2ebdb1860a9337c24 also assume they were issues with the same version (no clue what the "current" version was at the time, but whoever reported it was clearly running an older version).

PR #357, opened 2017-01-01
Make sign interaction be counted as a use

Phoenix616:
Since the change to the json formatting signs are able to include a clickEvent. ~~Currently you need to allow all Interactions with the interact flag to allow that but it fits right in with the other use flag cases.~~ EDIT: Apparently this is not the case with Vanilla sign events (they bypass all protections) and only a problem with how certain plugins check if signs can be used (via calling the [ProtectionQuery.testBlockInteract()](https://github.com/sk89q/WorldGuard/blob/master/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/ProtectionQuery.java#L88) method which returns false for signs as long as the full interact flag isn't set)

This also fixes the issue of allowing certain CraftBook sign mechanics in protected regions (like elevators).

### DIFF
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/util/Materials.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/util/Materials.java
@@ -806,6 +806,8 @@ public final class Materials {
             case JUNGLE_DOOR: return true;
             case ACACIA_DOOR: return true;
             case DARK_OAK_DOOR: return true;
+            case WALL_SIGN: return true;
+            case SIGN_POST: return true;
             default: return false;
         }
     }


wizjany:
I do not agree with this. This has been *intentionally* left out for a very long time mostly because other plugins should be handling this themselves depending on how they are using signs.

>Currently you need to allow all Interactions with the interact flag

Uh, what, no, source please? Region with no flags:
```
[00:07:08 INFO]: [WorldGuard] INTERACT SAND [CraftPlayer{name=wizjany}] @291,63,219 :PlayerInteractEvent [ALLOW]
[00:07:09 INFO]: [WorldGuard] INTERACT SIGN_POST [CraftPlayer{name=wizjany}] @291,64,219 :PlayerInteractEvent [ALLOW]
[00:07:10 INFO]: [WorldGuard] INTERACT WHITE_SHULKER_BOX [CraftPlayer{name=wizjany}] @ :PlayerInteractEvent [DENY]
```

There is a small case now for adding sign-interaction protection on the basis that signs can now have click actions in vanilla, but I'm not sure if it's really necessary.

At the least, there is a good argument *against* adding things simply because it would break setups that use signs where the **use** flag is not set to allow (because it wasn't needed before).

Phoenix616:
You are right, @wizjany, Vanilla sign click events currently bypass all protections, the problem currently only happens with plugins using the ProtectionQuery.testBlockInteract() method which forces you to set the full interact flag as you can't target only the stuff not caught by the other flags without setting them (use, chest-access, etc.). I still think that signs belong into the use category 'though as they can be used and adding the need to set a single flag isn't that bad in my opinion.

Alternatively it might be worth thinking about at least catching the click event via the interact flag.

PR #359, opened 2017-02-14
GPLv3 Markdown

DMP9-MC:
This PR rewrites the LICENCE file in markdown format.


wizjany:
why?

DMP9-MC:
Because markdown looks better and works nicely with GitHub

wizjany:
That seems pretty subjective, especially since github isn't even the only place people will look at it.
Also, as in sk89q/WorldEdit/pull/381, you forgot to actually look at the file you were replacing.

I am going to close these PRs because there doesn't seem to be any compelling reason to switch formats. There's really no advantage and no precedent.

PR #364, opened 2017-04-17
Add frosted-ice-melt and frosted-ice-form flags

Katrix:
Adds two new flags `frosted-ice-melt` and `frosted-ice-form`, to control the behavior of frosted ice. The melt flag does not prevent the cracks, only melting.

### DIFF
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
@@ -509,6 +509,13 @@ public class WorldGuardBlockListener implements Listener {
                     return;
                 }
             }
+
+            if (type == BlockID.FROSTED_ICE && wcfg.useRegions &&
+                    !plugin.getGlobalRegionManager().allows(
+                            DefaultFlag.FROSTED_ICE_FORM, event.getBlock().getLocation())) {
+                event.setCancelled(true);
+                return;
+            }
             return;
         }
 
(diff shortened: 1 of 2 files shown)


worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
```diff
@@ -509,6 +509,13 @@ public void onBlockForm(BlockFormEvent event) {
                     return;
                 }
             }
+
+            if (type == BlockID.FROSTED_ICE && wcfg.useRegions &&
+                    !plugin.getGlobalRegionManager().allows(
+                            DefaultFlag.FROSTED_ICE_FORM, event.getBlock().getLocation())) {
+                event.setCancelled(true);
+                return;
```
duchuyctlk: IMO `return` statement here is redundant. There is another `return` outside of `if` block.

Katrix-: I can remove it if that's desired. Added it as it was present in the same way a lot of other places.

DarkArc: Eh, it's fine as is, this is pretty common for the WorldGuard code base.

wizjany:
tbh i was hoping all the million flag suggestions would be implemented via the api or consolidated. most users already probably don't use the vast majority of flags provided by default.

worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
wizjany: yea it may be redundant now but these listener cases tend to get added to a lot so it may not be redundant in the future.

PR #365, opened 2017-05-01
Added blacklist to prevent lava from destroying blocks.

Ecconia:
Added a new list to the config "physics.disable-lava-damage-blocks" and extended the BlockFromToEvent listener with a check for this blacklist.
Added since it was not possible to prevent lava from destroying selected blocks, with the given configuration.


wizjany:
i tested this and noticed that the isAir condition (which is designed for things like this http://i.imgur.com/P9O30cJ.png where the liquid flows from above but the source gets cut off while it falls) doesn't properly differentiate the water and lava settings. not sure atm if we can get the previous state of the block? i assume this is also an issue with the current code though, but due to the lack of this setting it hasn't come up before.

Ecconia:
I also did thought about that, but didn't found a solution up to now. And yeah that also is an issue for the old code. It would have blocked lava which became air too.

Ecconia:
Obviously I thought about it after the PR... 
I can't see a way to see what the bock was before, or to detect the type of fluid from this Event. 
If I find some time I will do tests, in hope of answers. 
In case that there really is no way without a tracker to detect the type, it's possible to add a note to these two lists, that they can't detect which is the flowing one for these cases. The other way, would be to separate the air cases into another list, I don't think that is practical though. I would need to know all the cases where air is the flowing type.

Pieter12345:
@wizjany I would suggest to merge this PR since it adds a nice feature. This PR adds a lava flow blacklist which works exactly the same as the existing water flow blacklist. The "air" check problem can not be easily fixed due to how the Bukkit physics event works (No information about the disappeared lava/water block available). Using the same blocks in the water and lava blacklists results in a fluid blacklist for which this problem is not present, this feature is essential for the protection of for example redstone components from fluid.
When this PR will be merged, I suggest to update the documentation on the lava and water flow blacklists to explain the air problem and that it does not occur when a block is on both blacklists.

PR #367, opened 2017-05-14
Add support for Minecraft 1.12

md-5:
Minecraft 1.12 requires Java 8 and bundles Guava 21. This and the associated WorldEdit changes allow usage in that scenario. Unfortunately it's hard/impossible to avoid breakage as Mojang jumped over two years of Guava versions, which is longer than their deprecation cycle.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,7 +21,7 @@ buildscript {
 
     configurations.all {
         resolutionStrategy {
-            force 'com.google.guava:guava:17.0'
+            force 'com.google.guava:guava:21.0'
         }
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -26,8 +26,8 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.1'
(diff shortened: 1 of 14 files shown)


md-5:
Travis build would've failed because it can't depend on the accompanying WorldEdit PR

stuntguy3000:
Why is this still open?

wizjany:
because minecraft 1.12 wasn't released til yesterday and spigot wasn't released til 13 minutes ago. calm down.

LinearBotSystems:
Since I can't reach your issue tracker because the site is loading in an endless loop and I to say something remotely to this I will write it here. First my bug is I can't give rights like interact or use allow  in a second world created by multiverse 2.6 (WG 2.6.1 and newer) I don't know where the problem is WorldGuard, permissions or because Spigot since 1.12 don't copy the setting from the original wold anymore to the new.

To this topic I wanted to debug it myself first the list of external dependencies is absurd same goes of the list of function which are deprecated I would at least consider to make a forge with a 3.0 version which start with Spigot 1.12

wizjany:
1) the issue tracker is working fine
2) even if it isn't, there's a support forum

PR #368, opened 2017-05-27
Make flag list more readable

Blacksheep92:
Alphabetical sorting and color differentiation.

### DIFF
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/commands/region/RegionCommands.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/commands/region/RegionCommands.java
@@ -71,8 +71,6 @@ import org.bukkit.command.CommandSender;
 import org.bukkit.entity.Player;
 
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


wizjany:
originally the flags were given in an order grouped by their function, which made it easier to find a flag if you knew what kind of thing you were looking for but not the name. alphabetical order only helps to find a flag if you know the first part of the name already.

Blacksheep92:
I havn't noticed any meaningful groups in all the years of use.
Perhaps these groups should be name and better separated from each other?

wizjany:
kind of like this although over the years some things have been badly placed.

http://i.imgur.com/GKyWBHZ.png

you can see some distinctions between build flags, membership-based, mobs/explosions, natural growth, liquid/spread, etc etc.

as you said it's not completely obvious since they aren't separated or labeled, but they are there if you know to look for them.

Blacksheep92:
Ok, on closer inspection it makes sense.

My intention was to find a known flag faster. (But this could be solved by tab completion)

But the color separation is still a useful simplification:
http://i.imgur.com/6FitMJl.png

wizjany:
colors are nice, definitely an improvement over the previous block of red.
tab completion would also be nice lol.

Blacksheep92:
I've adapted the Pullrequest.
Only the color adjustments are included.

TomyLobo:
why is that even crammed into so few lines in the first place?
I think it should be one per line, with comments indicating the sections.

wizjany:
because this was written before minecraft had chat scrollback. there were probably also a lot less flags to worry about as well. and like i said the sections aren't really well defined.

TheMolkaPL:
This should be paginated imo. Also colors can be helpful to separate the sections - eg. red for building flags, green for spawning, etc.

wizjany:
well then make a PR @TheMolkaPL

wizjany:
merged, thanks.

PR #370, opened 2017-08-04
Update WorldGuardPlayerListener.java

aaomidi:
A new essentials (?) update broke the command blocking system.

This update is required to allow worldguard to block Essentials commands.


wizjany:
you realize highest priority means it gets it last right?

aaomidi:
Yes I do know what it means, no need to be salty :)

This ensures that this will be executed last, aka this will have the last say in what happens.

I'll just maintain a patch with this feature.

aaomidi:
@DarkArc I agree, they shouldn't.

Neither communities are budging on whats the right way to go (:

DarkArc:
That's unfortunate, however, WorldGuard is following best practices here, and as such, we won't be "fixing" this on our end; sorry.

wizjany:
having the last say is not what we want. we want the first say so we can cancel plugins that might be acting on the event later (even thought it's command *pre* process and not supposed to be acted on).

PR #372, opened 2017-10-27
Issue with plugin WorldGuard

MentosPL:
I have these plugins:
http://builds.enginehub.org/job/worldedit/9945/download/worldedit-bukkit-6.1.9-SNAPSHOT-dist.jar
http://builds.enginehub.org/job/worldguard/9934/download/worldguard-legacy-6.2.2-SNAPSHOT-dist.jar

And when i have more than 48 plugins WorldGuard don't load with error "there's no wolrdEdit" but WorldEdit run's propertly.

Plese someone, to download random 47 plugins, Put 46 plugins in folder plugins (With WorldEdit and Worldguard of course = [46 plugins + WG and WE = 48]), and run server using spigot 1.12.2. Everything is ok? disable server, and add the last plugin and run ;)


wizjany:
as noted in the last one, this is the wrong place to report issues. please stop.

PR #373, opened 2017-10-27
Why there is no issues tab?

MentosPL:
github.com/sk89q/WorldGuard/issues - not working ;-;


EvilOlaf:
Because the developers decided to not use the implemented issue tracker from Github. Quite simple.
http://dev.enginehub.org/youtrack/issues

PR #375, opened 2018-02-21
Add MapFlag

Andavin:
Signed-off-by: Mark <33119mark@gmail.com>

I found a use for a MapFlag in one of my projects I was working on. Just a useful utility for more complex flag setup.

### DIFF
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/MapFlag.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/MapFlag.java
@@ -1,3 +1,22 @@
+/*
+ * WorldGuard, a suite of tools for Minecraft
+ * Copyright (C) sk89q <http://www.sk89q.com>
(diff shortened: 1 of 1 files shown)


Andavin:
I don't understand. What about this would fail the build? The coding style? That's all I think that I gathered from the details.

gabizou:
It says in the log: 
```
[ant:checkstyle] /home/travis/build/sk89q/WorldGuard/worldguard-legacy/src/main/java/com/sk89q/worldguard/protection/flags/MapFlag.java:4:1: Disallowed import - org.apache.commons.lang.StringUtils.
```

me4502:
Does this still function correctly in the latest versions of WorldGuard? If so, I'll merge it.

Andavin:
This does indeed compile with the latest version of WorldGuard and as it is basically a small string parser there doesn't seem to be any other possibilities for version incompatibility.

me4502:
Thanks

PR #378, opened 2018-06-26
Add SNOWMAN_TRAILS flag

fabien91:
Hi,

I've add a new flag "SNOWMAN_TRAILS" because i want my players can use this function but not in my spawn.
This flag add the function to allow or not the trails left by the snowmen in a region.
I've see in the config that is possible to disabled in a world but it's an issue here, i've test this and it's working.

Fabien


me4502:
I'll be adding this in after the 1.13 release has been made, as this contains usages of deprecated APIs that are removed in WorldGuard 7.

PR #380, opened 2018-07-24
Fix for Spawner Eggs

frost-byte:
Add helper function in Materials to get related EntityType for spawner eggs
Fix onRightClick Handler when player uses a spawner egg
Added snakeyaml to gradle build for worldguard-core
Fix for ChunkHashTable Factory compile error


me4502:
This already exists, thanks anyway

frost-byte:
No problem, figured you'd knock it out on your own at some point.  Your solution was cleaner.

PR #385, opened 2018-08-20
Delay ProcessPlayerEvent by one tick on bukkit onEnable

Librazy:
This change allow other plugin to register their Handlers on existing player when /reload the server.

When /reload the server,  WorldGuard create Sessions for online players by firing ProcessPlayerEvent in onEnable. But other plugin that want to register their Handles is not enabled yet (because they need to depend on WorldGuard). So existing player will bypass other plugin's Handle for this session.

If we delay ProcessPlayerEvents by one tick, we can ensure all plugin are enabled and registered.


me4502:
Please fix your formatting and target the master branch

PR #389, opened 2018-10-05
Fix NPE in WorldGuardPlugin.wrapOfflinePlayer()

Brokkonaut:
The constructor of BukkitOfflinePlayer calls the constructor of BukkitPlayer with a null player. So without this fix it is not possible to create any BukkitOfflinePlayer.


LinzN:
@me4502

me4502:
Thanks

PR #393, opened 2019-01-25
[WIP] Allow NPCs to bypass region protections.

wizjany:
Plugins such as Citizens create fake Players on the server as NPCs. These NPCs trigger many of the same events that normal players do. Normally this would mean that you can even add the NPC (by UUID) to a region. However, when bypass permissions are checked, some permissions plugins (just LuckPerms? unsure of others) refuse to do permissions lookups on the main thread (since it requires UUID resolution?).
There is a config setting to allow this, but presumably(?) the lookup comes up false anyway (unless the admin actually assigned NPCs permissions).
It might be worth making a config setting for this option (in case NPCs should be restricted?), and double-checking what the actual behavior of LuckPerms is in these cases.

### DIFF
--- a/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/util/InteropUtils.java
+++ b/worldguard-legacy/src/main/java/com/sk89q/worldguard/bukkit/util/InteropUtils.java
@@ -70,6 +70,10 @@ public final class InteropUtils {
         UUID uuid = player.getUniqueId();
         String name = player.getName();
 
+        if (player.hasMetadata("NPC")) {
+            return true;
+        }
+
         if (uuid.equals(forgeFakePlayerUuid)) {
             return true;
         }


TomyLobo:
Do you think it makes sense giving the player control over the types of NPCs that are allowed?
Maybe they want citizens in, but not killbots or something :)

me4502:
NPC plugins don't really allow determining that, sadly

TomyLobo:
Luckily, there's Object.getClass() :)

wizjany:
Yea, it's not so much citizens-specific as they are the only major NPC plugin and they use it. Technically any other plugin could make NPCs their own way, mark entities with the same metadata string ("NPC") and it would be the same to us unless we start hooking directly into their API. This way is nice because it just uses the bukkit api.

Also from what I can tell, the base MC entity is not extended, but wrapped. So there's no way for us to get the "npc type" because the object is literally just a Player.

Sauilitired:
It extends CraftPlayer. No wrapping, except for internally: https://github.com/CitizensDev/Citizens2/blob/master/v1_13_R2/src/main/java/net/citizensnpcs/nms/v1_13_R2/entity/EntityHumanNPC.java

Either check the metadata (recommended) or check the absolute player class (not recommended)

wizjany:
except calling getClass() on the (Bukkit)Player doesn't reveal whether the "player" is a "citizen" or a "killbot".
the wrapping i was talking about is the "NPCHolder" part - the NPC object is connected to the actual entity, but not the same object. since (without using nms or citizens api) we only have access to the (Bukkit)Player, the "NPC" metadata is all we can see.

wizjany:
it just occurred to me that we still have a forge fake player override option from the days of mcpc/cauldron.
it might be worth just re-using that option instead of this.

me4502:
I very much agree with reusing that

PR #398, opened 2019-03-22
TimestampFlag

JOO200:
I want to use in several plugins the possibility to set timestamps for certain regions.
Is there a possibility to add a TimestampFlag?

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
@@ -1,0 +1,79 @@
+/*
+ * WorldGuard, a suite of tools for Minecraft
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldGuard team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldguard.protection.flags;
+
+import javax.annotation.Nullable;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+
+/**
+ * Stores a timestamp.
+ */
+public class TimestampFlag extends Flag<Date> {
+    private static final DateFormat DEFAULT_FORMAT = new SimpleDateFormat("yyyy-MM-dd,HH:mm:ss");
+
+    protected TimestampFlag(String name, @Nullable RegionGroup defaultGroup) {
+        super(name, defaultGroup);
+    }
+
+    protected TimestampFlag(String name) {
+        super(name);
+    }
+
+    @Override
+    public Date parseInput(FlagContext context) throws InvalidFlagFormat {
+        String input = context.getUserInput();
+        if("now".equalsIgnoreCase(input)) {
+            return new Date();
+        } else if("none".equalsIgnoreCase(input)) {
+            return null;
+        } else {
+            DateFormat format = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT, Locale.getDefault());
+            try {
+                return format.parse(input);
+            } catch (ParseException ignored) {
+                String pattern = format instanceof SimpleDateFormat ?
+                        ((SimpleDateFormat) format).toLocalizedPattern() :
+                        format.format(new Date());
+                throw new InvalidFlagFormat("Expected input in format: " + pattern);
+            }
+        }
+    }
+
+    @Override
+    public Date unmarshal(@Nullable Object o) {
+        if (o instanceof String) {
+            try {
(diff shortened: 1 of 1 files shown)


worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
```diff
@@ -0,0 +1,79 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldguard.protection.flags;
+
+import javax.annotation.Nullable;
+import java.text.DateFormat;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.Date;
```
wizjany: why not java 8 time? we don't have to be java 6/7 compliant anymore as was once the case!

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Locale;
+
+/**
+ * Stores a timestamp.
+ */
+public class TimestampFlag extends Flag<Date> {
+    private static final DateFormat DEFAULT_FORMAT = new SimpleDateFormat("yyyy-MM-dd,HH:mm:ss");
```
wizjany: any reason the default isn't something like iso 8601? this format seems...super non-standard? and susceptible to time zones?

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
```diff
@@ -0,0 +1,79 @@
...
+        super(name);
+    }
+
+    @Override
+    public Date parseInput(FlagContext context) throws InvalidFlagFormat {
+        String input = context.getUserInput();
+        if("now".equalsIgnoreCase(input)) {
+            return new Date();
+        } else if("none".equalsIgnoreCase(input)) {
+            return null;
```
wizjany: we have things like `/rg flag <region> blah` and `/rg flag <region>blah -e` to explicitly set null/empty values.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
```diff
@@ -0,0 +1,79 @@
...
+        } else {
+            DateFormat format = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.SHORT, Locale.getDefault());
```
wizjany: we should either be translating the user's time into utc (not easily accessible, unless passed through the context), or use utc by default.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
```diff
@@ -0,0 +1,79 @@
...
+                        ((SimpleDateFormat) format).toLocalizedPattern() :
+                        format.format(new Date());
+                throw new InvalidFlagFormat("Expected input in format: " + pattern);
+            }
+        }
+    }
+
+    @Override
+    public Date unmarshal(@Nullable Object o) {
+        if (o instanceof String) {
```
wizjany: should really be stored as a long in UTC, or at least ISO 8601.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
wizjany: also you write default, but it really should be `ONLY_ACCEPTED_FORMAT` by the look of it...

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
JOO200: LocalDateTime? I don't have problems with other time formats, even sql.Timestamp is fine.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
JOO200: I totaly agree with you. ISO8601 is better here. I'll change it.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
wizjany: http://lmgtfy.com/?q=java+8+time

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
JOO200: I copied that part from LocationFlag^^
I'll remove it

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
JOO200: Maybe use the server's time?
In most usecases the time from the server is in the same time zone as the player.

wizjany: 1) the server's time can change
2) I strongly disagree with this statement. Just because that's true for your (or some) server, doesn't mean it's always the case.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/TimestampFlag.java
wizjany: hm, should probably be removed from location flag too. that one is probably my fault.

JOO200:
@wizjany - all requested changes are made.

wizjany:
Merged manually. I fixed some formatting issues, and also made it possible to parse any TZ.
Previously, specifying input like `2019-03-30T13:00:00-04:00` would give 13;00 in UTC, ignoring the specified offset. Now, it parses to 17:00 UTC.

PR #402, opened 2019-04-28
Warn about consequences and add /stoplag confirm

TheMolkaPL:
Unexperienced server admins can accidentally erase all animals (sheep, pigs, cows) using `/stoplag`. This commit warns sender when executing `/stoplag` about its consequences and adds `/stoplag confirm`. The warning and confirmation applies only when `-c` flag is not present.


wizjany:
did you test this? i don't remember offhand if WG switched command managers to the intake one but `max = 0` won't work on the old one.

TheMolkaPL:
67ad427479201d285694873fab193ba3c11f7c30:

image

wizjany:
merged ac505f0064bf19e90375fa953eb1309e14d7b022
(I added a newline because I think it looks better)

PR #405, opened 2019-05-08
Enderpearl & chorus-teleport flags should be checked before handlers

isokissa3:
Before every handler would have been called, making them believe that the entry was successful


wizjany:
Merged 1ffa4b48b8448e216cf0ffc7f5c499c1710ba287

PR #406, opened 2019-05-08
Fix gitignore

Pieter12345:
Using Github Desktop on Windows, file paths with a leading slash do not seem to match anything. This commit changes the gitignore entries to a more widely supported format.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,17 +1,17 @@
-/.classpath
-/.project
-/.settings
-/target
+.classpath
+.project
+.settings/
+target/
 
 .DS_Store
 
 **/*.iml
 **/.idea
 
-/bin
-/dependency-reduced-pom.xml
+bin/
+dependency-reduced-pom.xml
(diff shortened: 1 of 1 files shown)


wizjany:
Wonder if this is an issue with github desktop, since intellij and git bash both seem to work. As long as this change is compatible it's probably fine though.

DarkArc:
It's slightly different, but the end effect should be about the same:
>A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but  not "mozilla-sha1/sha1.c".

>If the pattern ends with a slash, it is removed for the purpose of the following description, but it would only find a match with a directory. In other words, foo/ will match a directory foo and paths underneath it, but will not match a regular file or a symbolic link foo (this is consistent with the way how pathspec works in general in Git).

me4502:
This is what it should really be set to in the first place. Thanks

PR #408, opened 2019-05-11
Remove most usages of AsyncCommandHelper for new AsyncCommandBuilder.

wizjany:
Helper suffers from race conditions for short-lived tasks, leading to
some poor UX conditions such as errors not propagating to the user
(because the exception handler wasn't attached to the future yet), or
lack of success messages.

This commit replaces that system by a Builder which takes a callable to
begin, and then takes supervisor, delay message, and the success and
failure messages and handlers as parts of the builder. The success and
failure handlers wrap the callable itself before submitting to the
executor so they will always be run. The supervisor and delay are added
as listeners to the future since they aren't required if the task is
sufficiently short-lived (and to maintain compatibility with the classes
which are now in WorldEdit).


wizjany:
It's possible we want to move this into worldedit, though worldedit doesn't actually suffer from this condition unless you can submit a paste in less time than it takes to attach a listener to a future.

PR #410, opened 2019-05-20
New 1.14+ Logic

Xernium:
**I don't expect this to be merged**   
1.14.x saw some API changes, including the removal of the Bukkit Travel agent. I hope what I did here is somewhat of a suitable replacement. In any case this code needs to be looked at again.
Also, do we need a Sign/ Old type util class to check the new Sign types? I'm not so sure about those.

### DIFF
--- a/worldguard-legacy/build.gradle
+++ b/worldguard-legacy/build.gradle
@@ -10,7 +10,7 @@ repositories {
 
 dependencies {
     compile project(path: ':worldguard-core', configuration: 'shadow')
-    api 'org.bukkit:bukkit:1.13-R0.1-SNAPSHOT'
+    api 'org.bukkit:bukkit:1.14-R0.1-SNAPSHOT'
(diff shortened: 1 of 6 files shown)


Xernium:
Build check didn't complete- that wasn't the purpose as we aren't compiling against 1.14.x logic yet

wizjany:
Curious, did you test that portal change? Was chatting with some people in paper discord a bit ago and the issue seems to be that the EntityCreatePortalEvent is only called by the ender dragon portal, and the PortalCreateEvent gives a block list, but doesn't tell us *who* created the portal (for checking protection). Wondering if this was changed in spigot as a replacement?

Xernium:
@wizjany EntityCreatePortalEvent has the Entity that initiated it, I am running a test now as we speak. My computer is just slow, sorry

wizjany:
I'm aware it has an entity, but at least til 1.13, that entity was always a dying enderdragon afaik.

wizjany:
#411

PR #412, opened 2019-05-21
1.14+ Reinstate Nether-Portal Protection

Xernium:
This needs implementation from the side of other Entities too if that's not been done already


Xernium:
Fixes https://dev.enginehub.org/youtrack/issue/WORLDGUARD-4086

wizjany:
pretty sure 4086 is unrelated btw since that's the event that was used for the ender dragon portal

Xernium:
True. If there are other instances of it, but those needs replacing now anyway

PR #414, opened 2019-07-28
Add some Paper-specific features

wizjany:


worldguard-bukkit/src/main/java/com/sk89q/worldguard/util/profile/resolver/PaperProfileService.java
```diff
@@ -0,0 +1,57 @@
...
+import io.papermc.lib.PaperLib;
+import org.bukkit.Bukkit;
+
+import javax.annotation.Nullable;
+
+public final class PaperProfileService extends SingleRequestService {
+    private static final PaperProfileService INSTANCE = new PaperProfileService();
+
+    private PaperProfileService() {
+        if (!PaperLib.isPaper()) {
```
wizjany: note to self: this probably breaks atm

PR #419, opened 2019-12-16
Paper API 1.15

JOO200:
Use Paper API
bee nests can be modified by right click (like beehives)


wizjany:
oh, was literally about to do this. kewl thx.

PR #422, opened 2020-01-01
Update WorldGuard for CommandBook 3

DarkArc:
This must not be merged until the CommandBook builds are fixed, otherwise, we'll break the WorldGuard build.


wizjany:
note: figure out what to do with https://github.com/EngineHub/WorldGuard/blob/c413cf4123aabf4d4e989ee69e90f70a441acd51/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitConfigurationManager.java#L112
and probably update https://github.com/EngineHub/WorldGuard/blob/6c5319f355078c26569ac92dd351f26eb29a9458/worldguard-core/src/main/java/com/sk89q/worldguard/session/handler/GodMode.java#L45

PR #425, opened 2020-01-15
Allow set flags to be manipulate with `add` and `sub` keywords

DarkArc:
This adds support for updating an existing region, so that if you have a long list of items in a set flag, such as `deny-spawn` you can add or remove items, rather than overwrite.


wizjany:
Think I would rather have a set syntax like `+ item,...` and `- item,...` than the add/remove patterns. Works much better once we have piston set up with suggestions too, as +/- will appear at the top of the suggestion list before the possible values.

Also this should probably apply to MapFlag as well. Maybe we should abstract out AbstractCollectionFlag type?

PR #1562, opened 2020-04-25
Add teleport-message flag

stonar96:
Adds the ability to set a custom teleport message with the teleport-message region flag instead of the hard coded message up to now.

The default behavior (if the flag isn't set) matches the old behavior.

The message can be disabled by setting an empty message.


stonar96:
Sorry, I forgot to create a branch for this feature. I will close this PR and create a new one. See #1563.

PR #1563, opened 2020-04-25
Add teleport-message flag

stonar96:
Adds the ability to set a custom teleport message with the teleport-message region flag instead of the hard coded message up to now.

The default behavior (if the flag isn't set) matches the old behavior.

The message can be disabled by setting an empty message.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitPlayer.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitPlayer.java
@@ -25,6 +25,7 @@ import com.sk89q.worldedit.world.weather.WeatherType;
 import com.sk89q.worldedit.world.weather.WeatherTypes;
 import com.sk89q.worldguard.LocalPlayer;
 import com.sk89q.worldguard.WorldGuard;
+import com.sk89q.worldguard.util.MessagingUtil;
 import io.papermc.lib.PaperLib;
 import org.bukkit.BanList.Type;
 import org.bukkit.Bukkit;
@@ -190,9 +191,13 @@ public class BukkitPlayer extends com.sk89q.worldedit.bukkit.BukkitPlayer implem
         PaperLib.teleportAsync(getPlayer(), BukkitAdapter.adapt(location))
                 .thenApply(success -> {
                     if (success) {
-                        print(successMessage);
+                        if (successMessage != null && !successMessage.isEmpty()) {
+                            MessagingUtil.sendStringToChat(this, successMessage);
+                        }
                     } else {
-                        printError(failMessage);
+                        if (failMessage != null && !failMessage.isEmpty()) {
+                            printError(failMessage);
+                        }
                     }
                     return success;
                 });
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitPlayer.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitPlayer.java
@@ -191,13 +191,12 @@ public class BukkitPlayer extends com.sk89q.worldedit.bukkit.BukkitPlayer implem
         PaperLib.teleportAsync(getPlayer(), BukkitAdapter.adapt(location))
                 .thenApply(success -> {
                     if (success) {
-                        if (successMessage != null && !successMessage.isEmpty()) {
+                        // The success message can be cleared via flag
+                        if (!successMessage.isEmpty()) {
                             MessagingUtil.sendStringToChat(this, successMessage);
                         }
                     } else {
-                        if (failMessage != null && !failMessage.isEmpty()) {
-                            printError(failMessage);
-                        }
+                        printError(failMessage);
                     }
                     return success;
                 });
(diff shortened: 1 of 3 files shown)


worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
```diff
@@ -1111,8 +1111,10 @@ public void teleport(CommandContext args, Actor sender) throws CommandException
             }
         }
 
-        player.teleport(teleportLocation,
-                "Teleported you to the region '" + existing.getId() + "'.",
+        String message = existing.getFlag(Flags.TELE_MESSAGE);
+        message = message == null ? Flags.TELE_MESSAGE.getDefault() : message;
+        message = message == null ? "" : message.replace("%id%", existing.getId());
```
wizjany, line 1116: `getDefault` should return the compile-time default. is this nullcheck necessary?

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/Flags.java
```diff
@@ -134,6 +134,8 @@
     public static final LocationFlag TELE_LOC = register(new LocationFlag("teleport", RegionGroup.MEMBERS));
     public static final LocationFlag SPAWN_LOC = register(new LocationFlag("spawn", RegionGroup.MEMBERS));
 
+    public static final StringFlag TELE_MESSAGE = register(new StringFlag("teleport-message", "\u00A7dTeleported you to the region '%id%'."));
```
wizjany, line 137: Can you use the legacy serializer and text components as the other message flags do?

worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
stonar96: The second nullcheck is actually not necessary.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/Flags.java
stonar96: Of course.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitPlayer.java
```diff
@@ -190,9 +191,13 @@ public void teleport(Location location, String successMessage, String failMessag
...
                     if (success) {
-                        print(successMessage);
+                        if (successMessage != null && !successMessage.isEmpty()) {
+                            MessagingUtil.sendStringToChat(this, successMessage);
+                        }
                     } else {
-                        printError(failMessage);
+                        if (failMessage != null && !failMessage.isEmpty()) {
+                            printError(failMessage);
+                        }
```
DarkArc, line 200: I don't think this needs changed?

worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
```diff
@@ -1111,8 +1111,9 @@ public void teleport(CommandContext args, Actor sender) throws CommandException
             }
         }
 
+        String message = existing.getFlag(Flags.TELE_MESSAGE);
```
DarkArc, line 1114:
```
String message = existing.getFlag(Flags.TELE_MESSAGE);
if (message == null || message.isEmpty()) {
  message = Flags.TELE_MESSAGE.getDefault();
}

player.teleport(teleportLocation,  message.replace("%id%", existing.getId()), ...);
```

(For future reference, this is a pretty common pattern, get the thing, check if it's missing, assign it if it is)

stonar96: No, `isEmpty()` means no message.

_The message can be disabled by setting an empty message._

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitPlayer.java
stonar96: True it's not needed. This is just to be consistent. `MessagingUtil.sendStringToChat` is used on success, because it's configurable by a flag, the `failMessage` however isn't.

DarkArc: I see, `MessagingUtil` handles macro replacements, and since the message can be disabled you need the isEmpty check to avoid sending a blank line.

I'd just do this, the comment clarifies why it's there too, reducing the chance it's removed by accident in the future.
```
if (success) {
  // The success message can be cleared via flag
  if (successMessage.isEmpty()) {
    ...
  }
} else {
  printError(failMessage);
}
```

stonar96: Done.

worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
stonar96: I have replaced the ternary operator by an if statement and added commets to make everything clear. Done.

PR #1569, opened 2020-05-02
Create possibility to disable block fade from corals

JOO200:
https://minecraft.gamepedia.com/Coral_Block
Coral blocks will fade without water to dead coral blocks.
This should prevent coral blocks from fade to dead coral blocks


wizjany:
merged manually 
37ae70f

PR #1589, opened 2020-06-25
Update to Spigot 1.16.1 

JOO200:
Updated to 1.16.1
With the changed names from zombie pigmans to zombified piglins it's not easily possible to maintain backwards compatiblity to 1.15.

Updated PaperLib to 1.0.4 too.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/RegionProtectionListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/RegionProtectionListener.java
@@ -261,8 +261,8 @@ public class RegionProtectionListener extends AbstractListener {
                 canUse = query.testBuild(BukkitAdapter.adapt(target), associable, combine(event, Flags.CHEST_ACCESS));
                 what = "open that";
 
-            /* Beds */
-            } else if (Materials.isBed(type)) {
+            /* Beds or Respawn Anchor */
+            } else if (Materials.isBed(type) || type == Material.RESPAWN_ANCHOR) {
                 canUse = query.testBuild(BukkitAdapter.adapt(target), associable, combine(event, Flags.INTERACT, Flags.SLEEP));
                 what = "sleep";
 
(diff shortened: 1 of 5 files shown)


worldguard-bukkit/src/main/resources/plugin.yml
```diff
@@ -3,4 +3,4 @@ main: com.sk89q.worldguard.bukkit.WorldGuardPlugin
 version: "${internalVersion}"
 depend: [WorldEdit]
 softdepend: [CommandBook]
-api-version: 1.13
+api-version: 1.16
```
me4502, line 6: Hmm, never seen people up this beyond 1.13, does this actually do anything or is it just convention?

JOO200: that api-version 1.16 does make the plugin incompatible to 1.15 servers.

e.g. a 1.15 spigot/paper server doesn't load it
> [07:21:05 ERROR]: Could not load 'plugins\worldguard-bukkit-7.0.3-SNAPSHOT-dist.jar' in folder 'plugins'
> org.bukkit.plugin.InvalidPluginException: Unsupported API version 1.16

could be useful for users who misses the labels for the versions. With the Materials changes that plugin isn't compatible with 1.15 anymore.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/RegionProtectionListener.java
```diff
@@ -261,8 +261,8 @@ public void onUseBlock(final UseBlockEvent event) {
                 canUse = query.testBuild(BukkitAdapter.adapt(target), associable, combine(event, Flags.CHEST_ACCESS));
                 what = "open that";
 
-            /* Beds */
-            } else if (Materials.isBed(type)) {
+            /* Beds or Respawn Anchor */
+            } else if (Materials.isBed(type) || type == Material.RESPAWN_ANCHOR) {
```
JOO200, line 265: That's not the perfect solution for that.
Respawn Anchors has to be loaded by players with 4 glowstones. Only with full loaded anchors it's possible to set the spawn point.

And the message "sleep" doesn't match it. However I would let that inside the sleep flag because it's close enough to "set your spawnpoint"

wizjany:
rebased and merged

PR #1593, opened 2020-06-27
Add an option to disable bypass permission cache

TheMolkaPL:


wizjany:
I disagree with the premise of this PR. Did you see my response in discord? What problem are you attempting to solve here?

TheMolkaPL:
Yes, I saw it. I'm using LuckPerms which already caches permission checking and does proper invalidation when permissions are changed. I'm building a minigame server where different regions on maps (like spawns) are protected with WorldGuard. The server features an observer mode where players are able to fly around and they have bypass permission to bypass all WG's restrictions (like entry flags). After a player joins the game (from observer mode) they are able to bypass spawn protection for 2 seconds. Feel free to close this pull request, I'll then have to use my own fork, but I feel that this caching feature since it is not perfect should be switchable in the configuration.

wizjany:
Hm, that makes slightly more sense though:
1) you were asking about performance in discord, which seems to be completely irrelevant here (i.e. you're still allocating the tuple each check) and
2) sounds more like your spectators should belong to a group that's a member tbqh, this is kind of a misuse of bypass perms.
I'll give this more thought I guess.

TheMolkaPL:
> you were asking about performance in discord

Yeah, I didn't completely understand why this cache exists until I read description of https://github.com/EngineHub/WorldGuard/commit/83a7ed58a5482310a144a922880ffc0b75e5ccfc.

PR #1612, opened 2020-07-23
Added method to convert a ProtectedRegion to a WorldEdit's RegionSelector

JOO200:
Just a small addition to add a possibility to convert a ProtectedRegion to a RegionSelector.

I changed the command /rg select to the new method. It worked fine for my tests (cuboid, poly2d and global).

For my own plugins I want to mark a ProtectedRegion as selected sometimes. It could be a cool thing for new region types too.


wizjany:
This seems pretty weird because it's obviously leaking the goal (making selections) over what would otherwise be the correct api (convert to Region instead of Selector).
That is, if we didn't already have a clear goal of making selections when this API was written, it would be written to convert to a Region, which I think is correct.
That said, there obviously *is* a clear goal of making selections here. Additionally, it's a lot easier to get a Region from a Selector than vice versa.

JOO200:
I agree, it's weird to make a region through a region selector.
Maybe we can maintain both methods, one to create a region and one to create a region selector. Should not be too complicated and should match both cases.

wizjany:
after further consideration i'd rather have this in a separate util class.
coupling WE selectors/regions to the ProtectedRegion "interface" is pretty ugly and polluting imo, and implies there is some strict relationship.

PR #1618, opened 2020-07-30
Implement new alternative association mode to limit piston movement

JOO200:
Currently pistons can pull stuff from a lower prioritized region into their own region.
E.g. a plot owner can steal blocks from a city if he uses pistons for that.

After some hints from wizjany I managed to change the method for the lookup of the regions in the DelayedRegionOverlapAssociation class.

However I don't feel good with the current implementation. Without a config option this would be some breaking changes, with the config option it seems a little bit ugly and should be cleaned up.


JOO200:
I changed the requested things from the discord, should be ready for a review now

PR #1621, opened 2020-08-01
Optimized command usage for /rg list and /rg teleport

JOO200:
1. I added a flag to the teleport command to teleport the actor to the center of the region
e.g. `/rg teleport -c spawn` teleports the player to the center of the spawn region.
the player needs the permission `region.teleportoverride.spawn` for that.
With the permission the player will receive the teleport center options for regions without teleport flags for /rg list and /rg info.

2. I added a flag to the list command to filter the regions for special ids
This should help to filter regions based on their ids. I have 4000 regions in one world and sometimes there are special regions which i want to have listed. Currently i have to search in the list for that.
With the new flag I can search for those regions. E.g. `/rg list -i test` will list all regions containing `test` like `test-region` and `spawn-test`

There is an optional flag to switch to a regex mode (-r). `/rg list -r -i test.*\d` will match all regions with the regex pattern `test.*\d`, e.g. test-region-1 and testbanana25.
I made some test to verify the used pattern.

3. And I fixed some missing flags in the usage strings from the commands

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 dist: trusty
 jdk:
-  - oraclejdk8
+  - oraclejdk11
 notifications:
   email: false
 before_install: chmod +x gradlew
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,16 +1,22 @@
 # Changelog
 
-## 7.0.6 beta 1
+## 7.0.6
 
 * Add support for 1.17 blocks/items, drop support for 1.16.
 * Add use-dripleaf flag for "toggling" big dripleaf blocks.
 * Fix support for fully negative-height regions.
 * Fix crop-growth config option.
 * Be more rigorous with protecting blocks against dispenser behaviors.
+* Fix handling of lingering splash potion effects.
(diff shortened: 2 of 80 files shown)


wizjany:
1. -c has the potential to teleport you into something solid. not sure if there sure be a findFreePosition type thing there. minor gripe: teleportoverride has nothing to do with overriding. should probably just be named teleport-center or so, or teleport-without-flag
2. regex should be behind a permission. raw regex can be...abused. pattern should also be cached...tho potentially the jvm does this for us? can't remember details.
3. you forgot to add r flag to list usage text. otherwise fine tho eventually should get generated when we switch to piston.

JOO200:
I made the requested changes. I'm not happy with the findFreePosition method in AbstractPlayerActor. It teleports the player directly to the found location without any chance to verify the target location. I don't think it's helpful to teleport the player to a location which is not inside the region. Should I add some method inside WorldGuard to get a location inside the ProtectedRegion or are there any methods inside of WorldEdit?

At the moment center teleport is only executable in spectator gamemode (=no block bounds)

Issue #1630, opened 2020-08-24
Add a way to configure whether region bypass is on by default


bitrot-alpha:
**Is your feature request related to a problem? Please describe.**

I expect protection plugins to always protect by default, and right now, one mis-click means I need to recover blocks broken by myself or other admins. Whenever I log in with my admin account on my server, if I forget to turn off region bypass, I end up breaking something at the spawn when I'm not paying attention.

`basically add idiot protection for server admins who don't want to mess up player builds by accident`

**Describe the solution you'd like**

- **config.yml**
Add a line to set whether region bypass is ON or OFF by default for admins

**and/or**

- **Region flags**
Add a flag to set whether region bypass is automatically ON/OFF

- **Permissions**
A permission that can be denied to set region bypass to off by default, but not actually deny region bypassing altogether

- **Additional Idiot Protection**
A special message the first time you break a block in a protected WG region to run `/rg bypass` if it is off, and a warning if it is on.

**Describe alternatives you've considered**

Previously, I've used the Safezone feature from a Factions plugin, but have moved away from that because of the versatility and ease of use of WorldGuard regions. The one thing missing is that I don't have "idiot protection" like the Factions plugin did.

**Additional context**

_Note:_ I am using the `*` permissions node for my server's highest rank, Owner, and only denying permissions to that rank that break expected behavior/gameplay from plugins. I realize that there are _many_ worldguard permissions to set, so if I missed something, I'll happily close this issue if this feature already has an implementation.

_Kudos:_ I appreciate how well documented this family of plugins is. It only took me ten minutes to figure out that I needed to run `/rg bypass` to test if I was using WorldGuard properly.

wizjany:
0e73b50d6faf4dbb9566f87bea56165074b5e163

PR #1639, opened 2020-09-11
Add config option to deny region bypass by default

JOO200:
Should be some small thing.
The default value is set to "false" so there is no change to current behavior.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
@@ -280,7 +280,7 @@ public class BukkitWorldConfiguration extends YamlWorldConfiguration {
         maxClaimVolume = getInt("regions.max-claim-volume", 30000);
         claimOnlyInsideExistingRegions = getBoolean("regions.claim-only-inside-existing-regions", false);
         boundedLocationFlags = getBoolean("regions.location-flags-only-inside-regions", false);
-        disableDefaultBypass = getBoolean("regions.disable-default-bypass", false);
+        disableDefaultBypass = getBoolean("regions.disable-bypass-by-default", false);
 
         maxRegionCountPerPlayer = getInt("regions.max-region-count-per-player.default", 7);
         maxRegionCounts = new HashMap<>();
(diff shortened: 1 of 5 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
```diff
@@ -280,6 +280,7 @@ public void loadConfiguration() {
         maxClaimVolume = getInt("regions.max-claim-volume", 30000);
         claimOnlyInsideExistingRegions = getBoolean("regions.claim-only-inside-existing-regions", false);
         boundedLocationFlags = getBoolean("regions.location-flags-only-inside-regions", false);
+        disableDefaultBypass = getBoolean("regions.disable-default-bypass", false);
```
wizjany, line 283:
```suggestion
        disableDefaultBypass = getBoolean("regions.disable-bypass-by-default", false);
```

worldguard-core/src/main/java/com/sk89q/worldguard/session/Session.java
```diff
@@ -103,6 +103,8 @@ public void initialize(LocalPlayer player) {
 
         lastValid = location;
         lastRegionSet = set.getRegions();
+        disableBypass = WorldGuard.getInstance().getPlatform().getGlobalStateManager()
+                .get(player.getWorld()).disableDefaultBypass;
```
wizjany, line 108: Would like to have something here like `if (disableBypass) player.printInfo("You are bypassing region protection...")` so the user knows.

wizjany (changes requested):
Small UX changes requested.

Also "Closes #1630"

worldguard-core/src/main/java/com/sk89q/worldguard/session/Session.java
JOO200: DisableBypass means you're not bypassing region protections. Your suggested message doesn't make sense

And imo that shouldn't be necessary or another config option. On many servers there's already a spam at login from different plugins, I don't like the idea to add another message to the login spam

JOO200: I added a configurable message for users who can toggle bypass.

wizjany:
Er, just noticed this uses world config instead of global. Does it make sense for the login world to affect a server-wide (session) toggle? e.g. logging on to a world with the bypass off, then teleporting to one with it on, it will remain off. Think it's probably best to just hoist it to the server-wide config instead of per-world.

JOO200:
I agree, server wide config should be good enough. I changed it.

Maybe some world dependent stuff could be nice but that would need some listener stuff for world changes etc. I don't think that we need that atm.

wizjany (approved):

PR #1644, opened 2020-10-02
Change anvil usage to build permissions and add use-anvil flag

JOO200:
As discussed in the discord I changed the anvil usage to build permission level and added a new flag named use-anvil.


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/RegionProtectionListener.java
```diff
@@ -49,6 +49,7 @@
 import org.bukkit.ChatColor;
 import org.bukkit.Location;
 import org.bukkit.Material;
+import org.bukkit.Tag;
```
wizjany, line 52:
```suggestion
```

PR #1650, opened 2020-10-15
Add query methods to get values from MapFlags

JOO200:
I added some query methods to query value from MapFlags.
Usage:

- register the map flag (added in WorldGuard's Flags class):
```java
public static final MapFlag<ItemType, StateFlag.State> ITEM_PICKUP_TYPE =
            register(new MapFlag<>("item-material-pickup", new RegistryFlag<>(null, ItemType.REGISTRY), new StateFlag(null, true)));
```

- Use it in the code (changed RegionProtectionListener#onDestroyEntity for the test, just as test and should be cleaned up a little bit)
```java
        /* Item pickup */
        } else if (event.getEntity() instanceof Item) {
            ItemType itemType = BukkitAdapter.adapt(((Item)event.getEntity()).getItemStack()).getType();
            canDestroy = query.testBuild(BukkitAdapter.adapt(target), associable, Flags.ITEM_PICKUP_TYPE, itemType, Flags.ITEM_PICKUP, combine(event));
            what = "pick up items";

        /* Experience orb */
        } else if (event.getEntity() instanceof ExperienceOrb) {
```

- Compile and use the flags ingame

grafik

The regions are overlapping. Flag `item-pickup` is set to deny, so nobody can pickup items by default. The flag `item-material-pickup` is set on two regions:
- You can pickup dirt and anvils but you can't pickup stone.

The implementation does work with region inheritance and region priorities. It does search for the matched key with the highest priority.

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/protection/ApplicableRegionSet.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/protection/ApplicableRegionSet.java
@@ -144,6 +144,28 @@ public interface ApplicableRegionSet extends Iterable<ProtectedRegion> {
     <V, K> V queryMapValue(@Nullable RegionAssociable subject, MapFlag<K, V> flag, K key);
 
     /**
+     * Get the effective value for a key in a {@link MapFlag}. If there are multiple values
+     * (for example, if there are multiple regions with the same priority
+     * but with different farewell messages set, there would be multiple
+     * completing values), then the selected (or "winning") value will be undefined.
+     *
+     * <p>A subject can be provided that is used to determine whether the value
+     * of a flag on a particular region should be used. For example, if a
+     * flag's region group is set to {@link RegionGroup#MEMBERS} and the given
+     * subject is not a member, then the region would be skipped when
+     * querying that flag. If {@code null} is provided for the subject, then
+     * only flags that use {@link RegionGroup#ALL},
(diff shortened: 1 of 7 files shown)


JOO200:
Should be helpful for https://github.com/EngineHub/WorldGuard/pull/417

wizjany:
Think I'm going to have to double/triple-check the logic and docs here, as there's a lot of weird edge cases to cover. Possibly we should have some tests for this.
E.g., in cases where the `V` is a `State`, we should do a `queryState`-esque check where states are resolved against each other at equal priority. More generically, two of the same `K` that apply to a query with different values for `V` should use `Flag#chooseValue` to resolve `V`.

JOO200:
- I added some test cases to the PR. They should cover the basic logic things. Maybe I'll add some more tests.
- I changed the lookup process to use Flag#chooseValue instead of using the first available value.
- I added some sort of fallback flag to check for a Flag<Value> with same or higher priority.
E.g. the item-pickup flag could be set to deny on a higher prioritized region which will block item-pickups generally.

With the new methods in RegionQuery it should be easy to check for a MapFlag with a fallback flag. I added a method RegionQuery#testBuild for a method with MapFlag.

wizjany:
There's some lingering javadoc and formatting issues but the implementation *looks* complete, so I'm going to merge this now while I'm looking at stuff since who knows when I'll look next.
I'm sure the javadoc issues will get fixed...eventually...eh heh.

Issue #1475, opened 2019-01-06
Can't grow trees between WorldGuard Regions Owned By Same Player


LadyCailinBot:
WORLDGUARD-4025 - Reported by hutt132

Spigot 1.13.2
WorldGuard v7 b1767

I have it set up so players can claim individual chunks with WorldGuard.
However, if they try to grow a tree on the edge of their claim1, the tree will not grow over the region border to claim2, causing the tree to be cut in half.
**This happens when the two neighboring regions are owned by the same player.**

In the documentation it states "Tree growth grief (from outside a region) is prevented."
http://worldguard.enginehub.org/en/latest/regions/scope/

A few ways to resolve this issue would be:
1) Have WouldGuard check if the neighboring/touching regions are owned by the same player, and if so, allow tree growth across the borders.
2) Create an option to disable tree growth protection in the config.
3) Add a tree growth flag.

wizjany:
covered by #1763

PR #1661, opened 2020-10-27
Add use-owner-association option

stonar96:
This PR adds the option `use-owner-association` to the world configuration. I will explain this feature in the following example. Suppose you have a protected "plots" area in which plots are created for players. See picture below.

2020-10-27_20 35 02_edited

Players can now modify the surroundings (the protected "plots" area) by using pistons, fluids, ..., although they shouldn't be able to do this. If a player (Alice) owns two plots next to each other, it breaks pistons, fluids, ... between the plots. The plots are protected from outside though. See picture below.

The first issue can be fixed by setting `use-max-priority-association` to `true`. See picture below.

And the `use-owner-association` option can be used to get the desired behavior. See picture below.

I'd appreciate this feature. However, if you think this shouldn't be in WorldGuard, feel free to close it. I can easily maintain my own fork for this and I won't be upset. Thanks in advance.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
@@ -161,6 +161,7 @@ public class BukkitWorldConfiguration extends YamlWorldConfiguration {
         disableObsidianGenerators = getBoolean("protection.disable-obsidian-generators", false);
 
         useMaxPriorityAssociation = getBoolean("protection.use-max-priority-association", false);
+        useOwnerAssociation = getBoolean("protection.use-owner-association", false);
 
         blockPotions = new HashSet<>();
         for (String potionName : getStringList("gameplay.block-potions", null)) {
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/AbstractListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/AbstractListener.java
@@ -139,12 +139,13 @@ class AbstractListener implements Listener {
                 loc = entity.getLocation();
             }
             return new DelayedRegionOverlapAssociation(query, BukkitAdapter.adapt(loc),
-                    config.useMaxPriorityAssociation);
+                    config.useMaxPriorityAssociation, config.useOwnerAssociation);
         } else if (rootCause instanceof Block) {
             RegionQuery query = WorldGuard.getInstance().getPlatform().getRegionContainer().createQuery();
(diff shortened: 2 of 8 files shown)


wizjany:
https://github.com/EngineHub/WorldGuard/issues/1475#issuecomment-598218878

stonar96:
@wizjany Thanks for linking me to this comment. I had those scenarios in mind. However, I think it's more important that non player based mechanics are working as intended, rather than thinking about player based actions of co-owners and if they work or not. But you are right, either there should be only one (or no) owner per region (that's the case on my server), or one has to accept this inconsistency, if the `use-owner-association` is used. And for those to whom this applies it's a useful feature, which can be used or not.

By the way, this inconsistency isn't actually something that is introduced by this PR. Take for example the scenario in the second picture. The water can flow out of the region, however, a tree planted by an owner with bonemeal won't grow outside. So I would consider this as a general issue/consistency of WorldGuard, rather than one of this PR.

wizjany:
>  Take for example the scenario in the second picture above. The water can flow out of the region, however, a tree planted by an owner with bone meal won't grow outside.

huh? why is this inconsistent? a tree growing naturally over the border is treated the same as water flowing, while a tree grown by a player is treated the same as the player just placing a block across the border. the point is that a block source can't be an owner of two non-overlapping regions, while a player can. I don't think there's an inconsistency there.

stonar96:
> 
> 
> > Take for example the scenario in the second picture above. The water can flow out of the region, however, a tree planted by an owner with bone meal won't grow outside.
> 
> huh? why is this inconsistent? a tree growing naturally over the border is treated the same as water flowing, while a tree grown by a player is treated the same as the player just placing a block across the border. the point is that a block source can't be an owner of two non-overlapping regions, while a player can. I don't think there's an inconsistency there.

I know. Then let's not call it an inconsistency. I just wanted to point out that the "issue" described by the linked comment is not different than how WorldGuard already behaves in some cases. So actually I don't see an issue at all (and yes, I understand the scenario in the linked comment very well). It's up to you if this feature should be added or not. Thank's for replying.

stonar96:
Expressed differently: A sapling placed (not bone mealed) by a co-owner in a plot of Alice growing into another adjacent plot of Alice is by design of this PR and the `use-owner-association` option.

wizjany:
Yes, I'm aware of what the goal of this PR is. It just doesn't correctly (in my opinion) cover the case where "test2" is co-owned by alice and bob, as then alice and bob can grief test3 and test1 respectively. Essentially you're just targeting your *specific* use case (from a quick skim of your changes it looks like you don't even cover group domains?), so I linked my comment to provide more context on why things are as they currently are in *general*.

stonar96:
I have now covered group domains. (For more information, see the comments in the code.)

wizjany:
superceded by #1763

PR #1674, opened 2020-11-27
Add bamboo growth flag

### LINKED ISSUE #1673: Bamboo growth
The crop-growth flag does not prevent Bamboo from growing nor does any other flag for that matter.

HexedHero:
Fixes #1673 

https://i.imgur.com/7Flfef3.mp4


wizjany:
this is meant to be handled by crop-growth and all the machinery already exists in WG for it.
I have no idea why someone decided to throw spread events instead of growth events on the bukkit end, but their inconsistency doesn't mean we should handle it differently.

HexedHero:
Well its been like since Bamboo was added to the game and its not going to change, does that just mean WG will never support blocking Bamboo growth? The spread event makes sense since it spreads to another block, its the exact same as vines which is blocked in WG

wizjany:
guess the event from the javadoc:
```
 Examples:

    Wheat
    Sugar Cane
    Cactus
    Watermelon
    Pumpkin
    Turtle Egg
```
```
 Examples:

    Mushrooms spreading.
    Fire spreading. 
```
which one does bamboo fit in? 🤔

PR #1695, opened 2020-12-30
Move the Mob Blocker feature from CraftBook to WorldGuard

me4502:
This feature makes no sense within CraftBook, so I've made this PR to move it to WorldGuard.

Basically, it disables the MC functionality of letting mobs being able to enter vehicles. Ideally, it'd just block collision entry, but that isn't reliably possible with the Bukkit API.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardEntityListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardEntityListener.java
@@ -851,7 +851,7 @@ public class WorldGuardEntityListener implements Listener {
         ConfigurationManager cfg = WorldGuard.getInstance().getPlatform().getGlobalStateManager();
         WorldConfiguration wcfg = cfg.get(BukkitAdapter.adapt(event.getEntered().getWorld()));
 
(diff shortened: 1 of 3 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardEntityListener.java
```diff
@@ -845,6 +846,16 @@ public void onEntityChangeBlock(EntityChangeBlockEvent event) {
         }
     }
 
+    @EventHandler(priority = EventPriority.HIGH, ignoreCancelled = true)
+    public void onVehicleEnter(VehicleEnterEvent event) {
+        ConfigurationManager cfg = WorldGuard.getInstance().getPlatform().getGlobalStateManager();
+        WorldConfiguration wcfg = cfg.get(BukkitAdapter.adapt(event.getEntered().getWorld()));
+
+        if(wcfg.blockEntityVehicleEntry && !(event.getEntered() instanceof Player)) {
```
wizjany, line 854:
```suggestion
        if (wcfg.blockEntityVehicleEntry && !(event.getEntered() instanceof Player)) {
```

wizjany (approved):

PR #1700, opened 2021-01-03
Display CommandException if WorldEdit failed to initialize

Intelli:
This checks if WorldEdit is enabled, and resolves the NoClassDefFoundError that would normally show if WorldEdit failed to initialize.

An example of the error that this fixes can be seen here: https://pastebin.com/vy2gWj0w


wizjany:
This seems to be cosmetic? That is, we're not actually solving the problem here, we're just throwing a different exception in place of the NCDFE, but still getting the same result.

Intelli:
Yes, that's correct. This adds in the missing check alongside the existing instance checks in the getWorldEdit method, and throws a user friendly error using the established conventions.

wizjany:
Ok, just making sure (as per the title) you didn't think this was actually 'fixing' anything. A more user-friendly exception message is fine though.

PR #1714, opened 2021-01-17
added volume()-implementation for polygonal regions

alex9849:
I tried to implement a method that calculates the volume of a polygonal region. The problem is, that these polygonal coordinates always end in the bottom left corner of a block. What leads us to this problem:
https://puu.sh/H7Q4O/a2ca709e4c.png

In this case you could just enlarge all coordinates with one block, but sadly that approach doesn't work if the polygon gets more complex.
It's extremely inefficient, so I tried to parallelize it a bit and let the volume be cached, so it only gets calculated once.


wizjany:
as noted in discord, while this implementation may be fine for your needs, it's not going to fly with larger regions in more performance-sensitive areas. I would suggest you move this calculation to your plugin code in a static fashion instead of using `region#volume` for polys.

PR #1717, opened 2021-01-24
Add timings to third-party session handlers

wizjany:


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitConfigurationManager.java
```diff
@@ -56,6 +57,7 @@ public BukkitConfigurationManager(WorldGuardPlugin plugin) {
     public void load() {
         super.load();
         this.extraStats = getConfig().getBoolean("custom-metrics-charts", true);
+        this.extraTimings = getConfig().getBoolean("extra-timings", true);
```
JOO200, line 60: would change the options name to something better, e.g. "timings-for-session-handlers" or smth like that.

JOO200 (commented):
Only little detail is the options name, rest should be good.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitConfigurationManager.java
wizjany: I was thinking this could be re-used in case we want to add timings for other things as well in the future. Maybe I should make it a map in that case? eg
```yml
extra-timings:
    - session-handlers: true
```

PR #1739, opened 2021-03-07
Grab version from git

hannesa2:
Currently you set version with a commit, eg 7.0.5-SNAPSHOT but the last tag is 7.0.4 !
This means you make an optimistic tag preview with additional commit. This is confusing, error prone and makes afford.

Why not using git mechanism, where you base on latest commit and count the commits in between ?

With this pull request you switch to this


hannesa2:
Currently you set version with a commit, eg `7.0.5-SNAPSHOT`  but the last tag is `7.0.4` !
This means you make an optimistic tag preview with additional commit. This is confusing, error prone and makes afford.

Why not using git mechanism, where you base on latest commit and count the commits in between ?

With this pull request you switch to this

me4502:
I think you're misunderstanding the point of the git commit hash in the version tag?

hannesa2:
It's always a question of point of view, who misunderstand what

PR #1740, opened 2021-03-07
GitHub release

hannesa2:
It's based on #1739 

Simple set a tag and all is done !
Here you see a test example with artifacts https://github.com/MinecraftZock/WorldGuard/releases/tag/7.0.4.0

image


JOO200:
Current releases are on dev.bukkit.org, dev builds are on the build server.

Why should we change an existing and working system?
Btw: "Counting the commits" is no good choice as you can't check for wg builds from another branch.

hannesa2:
> Why should we change an existing and working system?

It changes nothing, it add something. 
With it it will increase trust, by a comprehensible build, instead of downloading from a site which I can't easy automate.
Btw, with this mechanism I will receive mails on new releases

mbax:
> Btw, with this mechanism I will receive mails on new releases

Doesn't BukkitDev offer that functionality still?

wizjany:
from a philosophical standpoint:
moving releases to github is bad for end users rn, as bad as CF/DBO is it's still more familiar for users. maintaining both is a nope.
as mentioned, things like emails for updates is already possible in CF/DBO.

from an impl standpoint:
changing our versioning scheme, making a release for the core jar (wut? this isn't even a maven repo lol?), debug printlns everywhere, breaking api compat (compile -> implementation)....

if you're serious about contributing I suggest you join discord to discuss your ideas in the dev team channel, but as it stands this PR strikes me as hacktoberfest-spam level changes.

mbax:
@hannesa2 [wrote](https://github.com/EngineHub/WorldGuard/pull/1740#issuecomment-793799118)

> kid/guys are forced to download dubious artifacts, with unclear source, from a infantile website as only source

I'm just going to address this point as it's the one I can best speak to.

BukkitDev has been *the* primary plugin source in this community for nearly a decade. While their review standards are a bit more opaque these days, the plugins are not instantly available and must pass a review first. It is a trusted source, and does not meet any "infantile" criteria I can think of.

PR #1743, opened 2021-03-25
Add command permissions for general, member and region commands

Rurla:
Useful for chat tips.


wizjany:
this breaks everything.

PR #1763, opened 2021-05-09
Add nonplayer-protection-domains flag

stonar96:
This PR is a generalization of #1661. The feature to remove boundaries for non player associables between regions can now be achieved by setting the `region-unions` flags of these regions (instead of ownership checks). The `region-unions` flag is a `Set<String>` flag, containing the names of the region unions in which the region is a member of. If a non player associable is a member of the region r1, then it is also member of the region r2, if there is at least one region union in which both regions are member of.


wizjany:
I think this is a generally good idea for resolving the issue, though I'd like to find time for a more in depth look at impl and some other minor things like the flag name before merging.

wizjany:
name suggestion: `nonplayer-protection-domains`
think it better describes the purpose of the flag

stonar96:
done

PR #1768, opened 2021-05-14
Fix Cause#isKnown()

stonar96:
Only the root cause should be checked. This will currently not change any gameplay behavior.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
@@ -90,26 +90,15 @@ public final class Cause {
     }
 
     /**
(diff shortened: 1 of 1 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
```diff
@@ -90,26 +90,15 @@ public boolean isIndirect() {
     }
 
     /**
-     * Return whether a cause is known. This method will return true if
-     * the list of causes is empty or the list of causes only contains
-     * objects that really are not root causes (i.e primed TNT).
+     * Return whether a cause is known. This method will return false if
+     * the list of causes is empty or the root cause is really not known
+     * (i.e primed TNT).
```
wizjany, line 95:
```suggestion
     * (e.g. primed TNT).
```

PR #1776, opened 2021-06-01
Fix Dispenser Shulker Box bug and cleanup

stonar96:


JOO200 (approved):

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
```diff
@@ -1161,7 +1161,7 @@ public void onCauldronLevelChange(CauldronLevelChangeEvent event) {
         }
 
         // handle water/lava placement
-        if (item != null && (item.getType() == Material.WATER_BUCKET || item.getType() == Material.LAVA_BUCKET)) {
+        if (item != null && (item.getType() == Material.WATER_BUCKET || item.getType() == Material.LAVA_BUCKET || Materials.isShulkerBox(item.getType()))) {
             Events.fireToCancel(event, new PlaceBlockEvent(event, cause, placed.getLocation(),
                     item.getType() == Material.WATER_BUCKET ? Material.WATER : Material.LAVA));
```
wizjany, line 1166: shulker isn't handled here, which is fine for our region protection but breaks stuff like blacklist :< (and potential third party handling, not that we're super-good about that anyway)

stonar96: Uh, yeah i forgot to add that here.

stonar96: I have added a check for dispensers, which can also be used for other special dispenser behavior in the future.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
```diff
@@ -1166,6 +1166,15 @@ public void onCauldronLevelChange(CauldronLevelChangeEvent event) {
...
             return;
         }
+
+        // handle special dispenser behavior
+        if (event instanceof BlockDispenseEvent) {
```
stonar96, line 1171: Alternatively we could use the `cause` instead of the `event` here, or move the whole thing to `onBlockDispense`.

wizjany: yea i'd just move this back to the onblockdispense handler directly. this method is just for various things shared between player and dispenser "interaction"

stonar96: done

PR #1779, opened 2021-06-04
Reduce CPU/allocs from Bukkit event listeners

PaulBGD:
Currently the system allocates and initializes a WorldEdit World instance in nearly every Bukkit event listener. This change bypasses the adapting and instead uses the Bukkit world, while still not exposing how internally the Bukkit world configuration is stored.

Here you can see the impact on CPU (collected with an async profiler to avoid the Java safepoint problem):

image

While admittedly the CPU usage is low (albeit on a 100 player server with a lowish amount of redstone) the impact on allocations is much higher:

image

From just the physics event, 20% of my server's allocations are adapting the world to look up the config. The result of this PR on my live server was GCs became roughly 20% less frequent, as well as a small increase in TPS.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitConfigurationManager.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitConfigurationManager.java
@@ -93,7 +93,14 @@ public class BukkitConfigurationManager extends YamlConfigurationManager {
      */
     @Override
     public BukkitWorldConfiguration get(World world) {
-        String worldName = world.getName();
+        return this.get(world.getName());
+    }
+
+    public BukkitWorldConfiguration get(org.bukkit.World world) {
+        return this.get(world.getName());
+    }
+
+    private BukkitWorldConfiguration get(String worldName) {
         BukkitWorldConfiguration config = worlds.get(worldName);
         BukkitWorldConfiguration newConfig = null;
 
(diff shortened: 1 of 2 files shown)


wizjany:
this likely should be moved up to abstractlistener but upon second look it seems half the listeners (including blocklistener here) don't extend it yet anyway.

PaulBGD:
If this PR won't be accepted without converting the listeners to AbstractListener, I can go ahead and convert them. Otherwise, is the implementation okay? I don't love the idea of casting the configuration manager, but otherwise I'm not too sure how to bypass the adaption.

PaulBGD:
Looking closer at the BukkitWorld impl in WorldEdit, it might be possible to optimize there. If we store the UUID of the world instead of creating a WeakReference, and then lazily get the world native access when it's needed it might end up being a cleaner improvement.

wizjany:
AbstractListener has access to the Plugin instance which could expose the Bukkit platform/config directly I suppose (internally at least - I don't consider anything outside of WorldGuard#wrapPlayer to be API, just as in WE only the adapter class is API).

I'd rather not not take this in the current state since it just adds tech debt for an extremely focused issue fix. If we're going to fix the world wrapping for config access I'd rather have it done consistently throughout the listeners instead of just in this single one which only shows up due to some lag machines spamming physics events or whatever the root cause is.

PaulBGD:
It's just redstone, but I agree. A more general fix would be better here, which I think might just need to be in WorldEdit. Given the BukkitWorld is designed to be tossed frequently throughout WE/WG, it's probably best to reduce the work it does directly.

PR #1780, opened 2021-06-05
WorldGuardBlockListener->AbstractListener with perf improvements

PaulBGD:
This brings the similar perf improvements from #1779 to all AbstractListeners, as well as converts the WorldGuardBlockListener to use the AbstractListener class.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
@@ -73,6 +73,8 @@ public class BukkitWorldConfiguration extends YamlWorldConfiguration {
     public boolean usePaperEntityOrigin;
     /* Configuration data end */
 
+    public boolean needsBlockListener;
+
     /**
      * Construct the object.
      *
@@ -309,6 +311,8 @@ public class BukkitWorldConfiguration extends YamlWorldConfiguration {
                 blockCreatureSpawn.add(creature);
             }
         }
+
+        this.needsBlockListener = this.noPhysicsGravel || this.noPhysicsSand || this.allowPortalAnywhere || this.ropeLadders;
 
         boolean useBlacklistAsWhitelist = getBoolean("blacklist.use-as-whitelist", false);
 
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldBlockListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldBlockListener.java
@@ -66,44 +66,15 @@ import org.bukkit.inventory.meta.ItemMeta;
  *
  * @author sk89q
  */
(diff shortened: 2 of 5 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
```diff
@@ -73,6 +73,8 @@
     public boolean usePaperEntityOrigin;
     /* Configuration data end */
 
+    public boolean needsBlockListener;
```
wizjany, line 76: name is rather inaccurate since it's just for physics, also should be @unreported

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldBlockListener.java
```diff
@@ -66,44 +66,15 @@
  *
  * @author sk89q
  */
-public class WorldGuardBlockListener implements Listener {
-
-    private WorldGuardPlugin plugin;
+public class WorldBlockListener extends AbstractListener {
```
wizjany, line 69: what? why rename?

wizjany (changes requested):
as mentioned in the previous PR the idea was to migrate all the listeners to abstractlistener so there's a unified way for listeners to retrieve the configuration.
you are obviously still more concerned for your one performance issue than doing this right. the random mis-named field addition kind of shows that you'd rather squeeze performance out that have it make sense.

if you're not up for that this will have to wait until I get the time for it. I'm not interested in adding more technical debt from third party contributions without some assurance they will get fixed, as was often the case in the past with WG.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldBlockListener.java
PaulBGD: The reason is to match the rest of the classes which extend AbstractListener. Only the old listeners which implement all the listener stuff themselves use the `WorldGuard....Listener` format, so I updated it to match the rest of the listeners.

PaulBGD:
You're right, I am mostly concerned with the performance improvement as it's my motivation to make this change at all. While I understand all the listeners need converting, I'm not sure why that would invalidate this PR to convert one of the listeners.

As for the 2nd commit I pushed to this branch, admittedly I pushed it to get opinions on whether there was any interest in that performance improvement. I am fine reverting the 2nd commit to this PR and sticking to just the conversion of the WorldBlockListener.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitWorldConfiguration.java
PaulBGD: https://github.com/EngineHub/WorldGuard/pull/1780#issuecomment-860891825

PaulBGD:
Do you want this rebased on the 1.17 branch?

PR #1784, opened 2021-06-11
WorldGuard update 1.17

JOO200:
- [x] Change dependencies temporarly to Spigot 1.17
- [x] Update other dependencies
- [x] Add new materials
- [ ] Add new flags for behavior changes
  - [ ] copper-fade flag for oxidation
  - [ ] smth against growing pointed_dripstones
- [ ] check for new copper behavior
  - [x] use an axe on a copper block as player
  - [ ] use an axe on a copper block through a dispenser
  - [x] use honeycombs on a copper block (convert it to a waxed block)
- [ ] Change spigot dependency to paper and uncomment Paper specific stuff

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 dist: trusty
 jdk:
-  - oraclejdk8
+  - oraclejdk11
 notifications:
   email: false
 before_install: chmod +x gradlew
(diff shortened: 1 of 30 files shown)


wizjany:
need to check for things like using an axe on a waxed copper block, both as a player and a dispenser

PR #1787, opened 2021-06-14
Add inheritance for all flags

stonar96:
Some flags are not queried by a location but obtained via a region id. The expected behavior is that inheritance still works for these flags. However, this is currently not the case and it is also not documented. This PR implements inheritance for all flags by using `FlagValueCalculator.getEffectiveFlag(ProtectedRegion, Flag, RegionAssociable)` instead of `ProtectedRegion#getFlag(T)`. The method `ProtectedRegion#getFlag(T)` should only be used inside `FlagValueCalculator` or for showing info about a region but never for behavioral decisions.

There are cases in which the code in this PR behaves differently than the current WG code. Thus this PR can be considered as a breaking change. However, these cases should really be very uncommon (e.g. membership affects the teleport flag in this PR but the default region group is everyone). The most critical change is that you need to be a member of a region to use `/rg tp -s <id>` by default because the default region group of the spawn flag is members.

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
@@ -52,6 +52,7 @@ import com.sk89q.worldguard.config.ConfigurationManager;
 import com.sk89q.worldguard.config.WorldConfiguration;
 import com.sk89q.worldguard.internal.permission.RegionPermissionModel;
 import com.sk89q.worldguard.protection.ApplicableRegionSet;
+import com.sk89q.worldguard.protection.FlagValueCalculator;
 import com.sk89q.worldguard.protection.flags.Flag;
 import com.sk89q.worldguard.protection.flags.FlagContext;
 import com.sk89q.worldguard.protection.flags.Flags;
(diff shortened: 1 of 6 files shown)


wizjany:
I'm honestly a bit on the fence with this one. It really goes both ways.

On one hand, it allows setting the teleport point for all "plots" at once by setting the flag on the "template" for those plots.

On the other hand, there's a potential issue where using `/rg tp x` doesn't teleport you to `x`, but to some parent of x which had higher priority, or a parent which the user is a member of, if they were not a member of x. The first (priority) is a well-documented caveat of combining inheritance and priority which is fine and probably doesn't happen much anyway.
The second is a bit of a weird one since previously it would return an error (something like "no teleport point for x" whereas now it will happily just teleport you to `some-parent` leading to confusion. Potentially if this is the case, the teleport message should give some sort of hint that `x` was not the exact region teleported to, but that might be hard to change in the message. 


On a side note, those static changes in FlagValueCalculator are binary incompatible. Didn't check but if those were already released as non-static in 7.0.5 I'd rather not break them. If it's just in snapshots that might be ok.

wizjany:
could you rebase and revert the static changes and add static overloads instead?

as for greeting/farewell those can't use effective values, they need to collect the message stack so WG doesn't repeat an outer region's message while moving to and from an inner region and other weird edge cases. it's really ugly and has a few issues right now, but using geteffective there is definitely worse.

stonar96:
Of course, I'll do that.

stonar96:
Done.

wizjany (approved):

PR #1790, opened 2021-06-30
Add member inheritance for non-player associables

stonar96:
There are a few reasons why I think this should be added:

1. As I have already described in EngineHub/WorldGuardDocs#24 non-player associables are also members of the parent regions. This essentially means that parent regions "inherit" members from child regions but not the other (if at all, expected) way around, as this is the case for player associables.
2. [FlagValueCalculator#getEffectiveFlag(ProtectedRegion, Flag<V>, RegionAssociable)](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/FlagValueCalculator.java#L470) implements a part of member inheritance by [collecting already seen (child) regions to pass](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/FlagValueCalculator.java#L512) them to [RegionAssociable#getAssociation(List<ProtectedRegion>)](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/association/RegionAssociable.java#L38). Of course this applies to player and non-player associables. This "partial" implementetion is even more confusing and unintuitive for the behavior in the case of non-player associables.
3. The other part of member inheritance for player associables is the [LocalPlayer#getAssociation(List<ProtectedRegion>)](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/LocalPlayer.java#L57) method, which internally also checks the parent regions of the passed regions via [ProtectedRegion#isOwner(LocalPlayer)](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/regions/ProtectedRegion.java#L282) for example, unlike the [AbstractRegionOverlapAssociation#getAssociation(List<ProtectedRegion>)](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/association/AbstractRegionOverlapAssociation.java#L87) method, which completely ignores the parent regions of the passed regions. Essentially this PR applies the same logic from player associables to non-player associables.

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/protection/association/AbstractRegionOverlapAssociation.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/protection/association/AbstractRegionOverlapAssociation.java
@@ -87,31 +87,35 @@ public abstract class AbstractRegionOverlapAssociation implements RegionAssociab
     public Association getAssociation(List<ProtectedRegion> regions) {
         checkNotNull(source);
         for (ProtectedRegion region : regions) {
-            if ((region.getId().equals(ProtectedRegion.GLOBAL_REGION) && source.isEmpty())) {
-                return Association.OWNER;
-            }
-
-            if (source.contains(region)) {
-                if (useMaxPriorityAssociation) {
-                    int priority = region.getPriority();
-                    if (priority == maxPriority) {
-                        return Association.OWNER;
-                    }
-                } else {
+            while (region != null) {
+                if ((region.getId().equals(ProtectedRegion.GLOBAL_REGION) && source.isEmpty())) {
                     return Association.OWNER;
                 }
-            }
 
-            Set<ProtectedRegion> source;
+                if (source.contains(region)) {
+                    if (useMaxPriorityAssociation) {
+                        int priority = region.getPriority();
+                        if (priority == maxPriority) {
+                            return Association.OWNER;
+                        }
+                    } else {
+                        return Association.OWNER;
+                    }
+                }
 
-            if (useMaxPriorityAssociation) {
-                source = maxPriorityRegions;
-            } else {
-                source = this.source;
-            }
+                Set<ProtectedRegion> source;
 
-            if (checkNonplayerProtectionDomains(source, region.getFlag(Flags.NONPLAYER_PROTECTION_DOMAINS))) {
-                return Association.OWNER;
+                if (useMaxPriorityAssociation) {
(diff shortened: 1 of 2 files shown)


wizjany:
Uh, based on the test alone without looking at any of the impl:

outer -> courtyard -> fountain
courtyard = 5, fountain = 10
courtyard is parent of fountain

a player in courtyard cannot build in fountain (correct behavior, right?)
a nonplayer in courtyard cannot build in fountain, but your change says they should be able to? what? Am i misreading something here.

stonar96:
> a player belonging to courtyard cannot build in fountain (correct behavior, right?)

No, a player belonging to courtyard can in fact build in fountain (member inheritance).
> a nonplayer in courtyard cannot build in fountain, but your change says they should be able to?

Yes, for the reason explained above.

stonar96:
I'll try to make a part of my initial message clearer.
Current behavior of WG
- for players:
  - Child regions inherit members from parent regions.
- for non-players:
  - Child regions do **not** inherit members from parent regions. (The reason is that there is no equivalent [iteration of parent regions](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/regions/ProtectedRegion.java#L290) for non-players. That's what this PR implements.)
  - Parent regions "inherit" members from child regions. (The reason is that [this](https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/association/DelayedRegionOverlapAssociation.java#L70-L71) also returns parent regions and not only the regions that are physically at that location.)

Due to the last bullet point (which is **not** fixed by this PR), the behavior will still not be completely correct, however, it will be more correct than it was before. Note that the last bullet point is actually breaks region protection because due to this players can modify parent regions (e.g. with pistons) in which they can't build. I will probably open another PR to fix this later.

wizjany:
oh wait yea, i totally forgot that courtyard is a parent of fountain despite writing it out.
so this allows NPAs in courtyard to build in fountain, since members of courtyard should be inherited members of fountain (like players), but if fountain is higher prio, a NPA in fountain can still be restricted from building in courtyard via use-max-prio.
i'm glad I wrote those tests tbh, trying to understand the algorithms and your description takes too much brain power.

PR #1814, opened 2021-08-13
Fix redefinition of parent regions

stonar96:


worldguard-core/src/main/java/com/sk89q/worldguard/protection/managers/index/HashMapIndex.java
```diff
@@ -74,9 +77,13 @@ private void performAdd(ProtectedRegion region) {
 
             ProtectedRegion existing = regions.get(normalId);
 
-            // Casing / form of ID has changed
-            if (existing != null && !existing.getId().equals(region.getId())) {
-                removed.add(existing);
+            if (existing != null) {
+                removeAndReplaceParents(existing.getId(), RemovalStrategy.UNSET_PARENT_IN_CHILDREN, region, false);
```
wizjany, line 81: why is rebuildIndex false? if the region's bounds have changed, doesn't that mean the prtree and chunk table indices need to rebuild?

stonar96: Yes, that's done by the methods that call performAdd.

PR #1841, opened 2021-10-02
blockTNTExplosions = true not longer canceles the explosion event

sidey383:


wizjany:
not sure what the idea here is, please elaborate if you think there's a bug somehow?

sidey383:
If I set block-TNT-Explosions to true explosions will be canceled.
I think this is a very strange behavior.

wizjany:
it's a verb. as in "prevent".

PR #1857, opened 2021-11-30
Add /rg migrateheights.

### LINKED ISSUE #1856: Command to migrate regions at vertical heights
### The Problem

As of 1.18, the world build limits go from -64 to 320 or something. Existing regions that were previously from bedrock-to-build-limit (e.g. via `//expand vert` or similar) will no longer cover the full world height. This may cause an unintentional lapse in protection for servers where fully vertical regions are important for whatever reason.

### A Solution

A command such as `/rg migrateheights [world]` which will, for every region that current goes from 0-255, expand it to go from -64 to 320.

### Alternatives

* Have people manually select, expand, redefine each region
* Have some external script that can migrate a region file (or sql loljk)

### Anything Else?

Dropping this here cuz I've meant to look into it several times this week but kept forgetting.

wizjany:
Closes #1856.

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/protection/managers/migration/WorldHeightMigration.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/protection/managers/migration/WorldHeightMigration.java
@@ -1,0 +1,114 @@
+/*
+ * WorldGuard, a suite of tools for Minecraft
+ * Copyright (C) sk89q <http://www.sk89q.com>
+ * Copyright (C) WorldGuard team and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU Lesser General Public License as published by the
+ * Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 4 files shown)


worldguard-core/src/main/java/com/sk89q/worldguard/protection/managers/migration/WorldHeightMigration.java
```diff
@@ -0,0 +1,108 @@
...
+
+        int min = -64;
+        int max = 319;
+        World world = WorldGuard.getInstance().getPlatform().getMatcher().getWorldByName(store.getName());
+        if (world != null) {
+            min = world.getMinY();
+            max = world.getMaxY();
+        }
+        for (ProtectedRegion region : regions) {
+            if (region.getMinimumPoint().getBlockY() == 0
```
JOO200, line 87: Maybe change it from min=0 and max=255 to min<0 and max>255 for people with datapacks in 1.17.
Or stupid developers (such as me) who created regions from 0...256

PR #1862, opened 2021-12-23
Protection flags for lecterns

Sacryn:
How to configure use/interact/chest-access to allow for players to read books on lecterns without being able to take them seem to be a topic that pops up every now and then. Current solution seem to revolve around using multiple nested regions if you want people to access chests, but not be able to take books from lecterns.

My solution was to move lectern access to separate flags from chest access, thus eliminating the need for multiple nested regions.


wizjany:
This is way over-complicating things. Taking items is chest access. Using the lectern is using. There is literally an FAQ on our docs that covers this. For most people you only need one flag since chest access generally isn't wanted for protected regions. We don't need two more flags for something like this.

PR #1866, opened 2021-12-29
Add custom Domains for domains used through the api

Joo200:
Sometimes it can be handy to add custom domains instead of the player uuid, name or a permission group.

E.g. if you have more dynamic domains or you don't want to sync the members and/or owners between your region and worldguard.

### Idea

Add some own domains to the DefaultDomain of regions. E.g. add your custom domain which determines the membership status by playtime.

You add your "ApiDomain" (maybe we can find a better name here) to the DefaultDomain with some informations how to parse your custom Domain to the regions.yml and backwards.  There is a Registry for custom Factories for ApiDomains, where you have to register the ApiDomain and the Factory in onLoad (like the existing FlagFactory).

### Missing points

- [find better names:
  - [x]  ApiDomain --> maybe CustomDomain?
- [x] Display the names of the custom ApiDomains (or domain.toString()) in /rg info
- [x] ~~Add this to sql storage (or ignore it)~~ we just ignore it, sql storage can't handle custom domains.
- [x] Check the usage to fit the needs
- [x] Check whether the ApiDomain storage does work fine
- [x] Code cleanup
- [x] Documentation


wizjany:
initial thoughts:
1) CustomDomain > ApiDomain
2) there's no support for command usage or parsing yet, which means domains have to be added to regions by api. we should ultimately allow for `/rg addmember <id> mydomain:somevalue` in command usage. (wonder if it might be useful to have a perm for which domains can be used similar to the way flags perms work?)
3) following 2, `:` should not be a valid character in domain names as that should split the name and value when parsing

JOO200:
I looked in the code and I don't think we should add the custom domains to /rg addmember|addowner|removemember|removeowner commands. With the current DomainInputResolver there is no way to check for actors permissions and there is not really a "context" for the CustomDomain (e.g. the actor, the used ProtectedRegion or whether it's the owner or member domain).

Two new commands should work better for that: 
`/rg ownerdomains <region> <domainname> [input]` and `/rg memberdomains <region> <domainname> [input]`
Command syntax is like the `/rg flag <region> <flag> [input]`.

JOO200:
I removed the new commands as discussed on the discord.
After some tests it looks fine for me, it works fine for me.

Maybe there should be some way to get a list of available custom domains but I don't know how that should be done.
e.g. `/rg domains`? or `/rg addmembers/addowners/delmembers/delowners`?

wizjany:
not super-worried about discoverability atm as these will probably be niche for a while and will never be as much as flags. once piston happens suggestions will help.

worldguard-core/src/main/java/com/sk89q/worldguard/domains/CustomDomain.java
```diff
@@ -0,0 +1,108 @@
...
+import com.sk89q.worldguard.domains.registry.CustomDomainContext;
+import com.sk89q.worldguard.domains.registry.InvalidDomainFormat;
+import com.sk89q.worldguard.util.ChangeTracked;
+
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+public abstract class CustomDomain implements Domain, ChangeTracked {
+    private static final Pattern VALID_NAME = Pattern.compile("^[A-Za-z0-9\\-]{1,40}$");
```
stonar96, line 32: Upper case characters don't work properly. E.g. using `/rg addmember` doesn't find the registered `CustomDomain`.

worldguard-core/src/main/java/com/sk89q/worldguard/domains/DefaultDomain.java
```diff
@@ -257,7 +326,7 @@ public boolean contains(String playerName) {
 
     @Override
     public int size() {
-        return groupDomain.size() + playerDomain.size();
+        return groupDomain.size() + playerDomain.size() + customDomains.size();
```
stonar96, line 329: Shouldn't this call `size()` for each element in `customDomains`?

worldguard-core/src/main/java/com/sk89q/worldguard/domains/registry/SimpleDomainRegistry.java
```diff
@@ -0,0 +1,167 @@
...
+
+    public void setInitialized(boolean initialized) {
+        this.initialized = initialized;
+    }
+
+    @Override
+    public void register(String name, DomainFactory<?> domain) throws DomainConflictException {
+        synchronized (lock) {
+            if (initialized) {
+                throw new IllegalStateException("New flags cannot be registered at this time");
```
stonar96, line 57: flags -> domains

worldguard-core/src/main/java/com/sk89q/worldguard/domains/registry/InvalidDomainFormat.java
```diff
@@ -0,0 +1,28 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License
+ * for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldguard.domains.registry;
+
+public class InvalidDomainFormat extends Exception {
```
stonar96, line 22: `InvalidDomainFormat` -> `InvalidDomainFormatException`

JOO200: Disagree, InvalidDomainFormat has the same format as InvalidFlagFormat
https://github.com/EngineHub/WorldGuard/blob/master/worldguard-core/src/main/java/com/sk89q/worldguard/protection/flags/InvalidFlagFormat.java#L22

wizjany: unfortunately this naming dates back to 2011 and the original addition of region flags. technically a breaking change to rename it now though. maybe can deprecate and migrate it (easily via extension) for the future if we want.

worldguard-core/src/main/java/com/sk89q/worldguard/domains/DefaultDomain.java
JOO200: I'm not completly sure how to handle this. Technically an empty custom domain "is there" and could change stuff. So I guess this implementation should be fine.
We can change it later if there are any issues.

worldguard-core/src/main/java/com/sk89q/worldguard/domains/registry/InvalidDomainFormat.java
JOO200: I renamed InvalidFlagFormat to InvalidFlagFormatException with full backwards compatibility.
And InvalidDomainFormat is now InvalidDomainFormatException.

wizjany (approved):

PR #1868, opened 2022-01-01
Optimize Region List and Region Teleport commands

JOO200:
After I was unable to fix git for https://github.com/EngineHub/WorldGuard/pull/1621 this is the new PR.
I removed the regex pattern matching as I don't see a good way to implement it. Storing the pattern doesn't work that fine and it's enough to have a flag to filter for region ids containing a string.

The changes are pretty straight forward and should improve the user experience.

Optimized command usage for /rg list and /rg teleport:
* added a flag to the teleport command to teleport the actor to the center of the region
* added a flag to the list command to filter the regions for special ids
* ~~added a flag to the list command to change the filter variant to regex~~
* added a flag to the list command to get regions intersecting your selection

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
@@ -30,6 +30,7 @@ import com.sk89q.worldedit.WorldEdit;
 import com.sk89q.worldedit.command.util.AsyncCommandBuilder;
 import com.sk89q.worldedit.extension.platform.Actor;
 import com.sk89q.worldedit.extension.platform.Capability;
+import com.sk89q.worldedit.regions.Region;
 import com.sk89q.worldedit.util.Location;
 import com.sk89q.worldedit.util.formatting.component.ErrorFormat;
 import com.sk89q.worldedit.util.formatting.component.LabelFormat;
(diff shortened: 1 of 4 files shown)


wizjany:
regex matching is gone but remains in your description, cmd flags, and import.

spectator requirement still ugly but idk what else to do without a lot more work either so ¯\\\_(ツ)_/¯

worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
```diff
@@ -475,6 +479,16 @@ public void list(CommandContext args, Actor sender) throws CommandException {
...
         }
 
+        if (args.hasFlag('s')) {
+            ProtectedRegion existing = checkRegionFromSelection(sender, "tmp");
+            task.filterByIntersecting(existing);
+        }
+
+        // -i string is in region id
+        if (args.hasFlag('i')) {
+            task.filterIdByMatch(idFilter = args.getFlag('i'));
```
wizjany, line 489: `idFilter` is assigned here but not used anywhere? could just drop the local entirely?

worldguard-core/src/main/java/com/sk89q/worldguard/commands/task/RegionLister.java
```diff
@@ -144,20 +151,29 @@ public boolean isContainedWithin(DefaultDomain domain) throws CommandException {
...
     }
 
+    public void filterIdByMatch(String idFilter) {
+        this.idFilter = idFilter;
+    }
+
     @Override
     public Integer call() throws Exception {
         Map<String, ProtectedRegion> regions = manager.getRegions();
+        Collection<ProtectedRegion> iterableRegions = manager.getRegions().values();
```
wizjany, line 161: should use `regions` again, `getRegions` is a full copy so this is both a performance and race condition.

wizjany (changes requested):
just tested this and noticed the regionlistbox produced hasn't been updated to include the new matchers in the command, so paging through will give different results than the initial list.

wizjany (approved):

PR #1917, opened 2022-05-29
Fix volume calculation

abbysrc:
This commit reintroduces polygonal volume calculation. Full explanation of this can be found here: https://stackoverflow.com/a/2263571

Bear in mind that I am an idiot and have probably made a stupid mistake.


wizjany:
is this a serious attempt or did you just blindly copy something off stackoverflow and not even look at what you were doing?

abbysrc:
In all honesty, most likely the second option. I have a feeling I was half asleep. You could've phrased the response a bit less aggressively though.

wizjany:
i mean, it's using three entirely undeclared variables. completely ignoring the mathematical issue, this is not even valid code. if it were another time of year i'd call this hacktoberfest spam.

abbysrc:
Point taken.

PR #1940, opened 2022-07-25
Introduce uninitialize for sessions

aromaa:
Introduces uninitialize for sessions that is called when the player disconnects or the server is shut down. Currently handlers don't get any notifications when the mentioned events occur and are unable to restore any player state they are holding to.

For example, this can be seen when entering to a region in survival with a `game-mode creative` flag, disconnecting and waiting for the session to timeout. When connecting back to the server and walking out of the region you can see that the player is not switched back to survival as they normally would be.

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/session/handler/FlagValueChangeHandler.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/session/handler/FlagValueChangeHandler.java
@@ -47,7 +47,7 @@ public abstract class FlagValueChangeHandler<T> extends Handler {
 
     @Override
     public final void uninitialize(LocalPlayer player, Location current, ApplicableRegionSet set) {
-        onAbsentValue(player, current, current, set, lastValue, MoveType.OTHER_NON_CANCELLABLE);
+        onClearValue(player, set);
         lastValue = null;
     }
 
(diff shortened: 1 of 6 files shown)


worldguard-core/src/main/java/com/sk89q/worldguard/session/handler/FlagValueChangeHandler.java
```diff
@@ -45,6 +45,12 @@ public final void initialize(LocalPlayer player, Location current, ApplicableReg
         onInitialValue(player, set, lastValue);
     }
 
+    @Override
+    public final void uninitialize(LocalPlayer player, Location current, ApplicableRegionSet set) {
+        onAbsentValue(player, current, current, set, lastValue, MoveType.OTHER_NON_CANCELLABLE);
```
aromaa, line 50: Not the most beautiful thing. We could also add a method like `onClearValue(LocalPlayer, ApplicableRegionSet)` for "completeness" where the default implementation would call back to `onAbsentValue` to not break existing plugins.

wizjany: I definitely prefer having a dedicated clear, gives more flexibility to things that need it for additional cleanup etc.

Issue #1963, opened 2022-11-05
WorldGuard Report doesn't show Paper server with Velocity Proxy


Joo200:
### WorldEdit Version

7.2.12

### WorldGuard Version

7.0.8

### Platform Version

Paper 1.19.2

### Confirmations

- [X] I am using the most recent Minecraft release.
- [X] I am using a version of WorldEdit compatible with my Minecraft version.
- [X] I am using a version of WorldGuard compatible with my Minecraft version.
- [X] I am using the latest or recommended version of my platform software.
- [X] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [X] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

Using Velocity as Proxy is not correctly displayed in the WorldGuard report:

```
Online Mode:
	enabled?: false
	BungeeCord support?: false
	Velocity support?: false
```

### Expected Behavior

Display this correctly

### Reproduction Steps

1. `/wg report -p` on a server with Velocity support enabled

### Anything Else?

Paper changed the config stuff in 1.19.2.
This is currently not fixable, because the config is currently not accessible.

ToDo: Add a PR to paper to allow `Bukkit.isProxyOnlineMode()` or smth like that by:
`io.papermc.paper.configuration.GlobalConfiguration.get().proxies.isProxyOnlineMode()`

PR #1971, opened 2022-12-09
Update to Minecraft 1.19.3

JOO200:
Changes

- Updated dependencies
- Updated Gradle to 7.6
- Added new Spawn Egg types to Materials
- Switched to 1.19.3 Tags for Signs (this will be a breaking change, ALL_SIGNS tag was not available in 1.19.2
- Fixed https://github.com/EngineHub/WorldGuard/issues/1963
- Bookshelfs are now modified by rightclick

ToDo:

- [ ] Add all the stuff from the 1.20 preview (Bamboo). They might be already handled with tags but are not contained in the tags now.
- [ ] Check for other changed stuff


wizjany:
1) this should point at 7.x i think?
2) any way we can fall back with Bukkit.getTag or something? was really hoping there wouldn't be breaking changes here

JOO200:
1. didn't change the version. maybe we should set version/7.0.x as default branch in github?
2. yes, i added some issue handling for 1.19-1.19.2. Those changes shouldn't be breaking now.

wizjany (approved):

PR #1988, opened 2023-03-19
Improve NPC handling in WorldGuard's listeners

JOO200:
This PR should minimize errors based on wrong interpretation of Entities of type Player from NPC plugins such as Citizens

- Do not listen to events from NPCs such as Citizens.
- Don't handle NPCs as players.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
@@ -135,7 +135,7 @@ public final class Cause {
     @Nullable
     public Player getFirstPlayer() {
         for (Object object : causes) {
-            if (object instanceof Player p && !p.hasMetadata("NPC")) {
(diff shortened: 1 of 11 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
```diff
@@ -134,8 +135,8 @@ public Object getRootCause() {
     @Nullable
     public Player getFirstPlayer() {
         for (Object object : causes) {
-            if (object instanceof Player) {
-                return (Player) object;
+            if (object instanceof Player p && !p.hasMetadata("NPC")) {
```
wizjany, line 138: any reason this one checks metadata directly instead of using the util method?

JOO200: No, i started the changes with individual checks for the metadata and used the Entities.isNPC() later.

This was an old relict from that. I changed some other metadata checks to that method too.

wizjany (approved):

PR #2011, opened 2023-06-08
Add 1.20 Support to WorldGuard

Joo200:
After the snakeyml fix we should add full 1.20 support.

There are no breaking changes so we can keep the compatibility with 1.19.

# ToDos:

- [x] Add sniffer egg protection as config option or merge it with turtle-egg's protection.
- [x] Validate correct protection for Suspicious sand and gravel (can be modified by brushes)
- [x] Evaluate the protection for Chiseled Bookshelf. Maybe move the protection to chest-access flag (sort of inventory?)
- [x] Protect usage Waxing of Signs
- [x] Protect text modifications on unwaxed signs


Brokkonaut:
Signs should be protected against editing if someone cannot build at the sign location

Joo200:
> Signs should be protected against editing if someone cannot build at the sign location

Currently signs should be protected by the use flag. However I agree that changing signs should be considered building.

I'll add it to the ToDo List

me4502 (approved):

PR #2012, opened 2023-06-08
Add Folia Support;

HarvelsX:
Adds basic support for Folia with the new scheduler system.
## Requirements Checklist
+ [ ] Create an adapter interface for new schedulers and implementations.
+ [ ] Implement region-specific, entity-specific task runs.
+ [ ] Rework `SchedulerReport`, but currently disable it for Folia;


me4502:
Does this retain Spigot support, as well as actually work given WorldEdit isn't marked as supporting Folia?

Kadeluxe:
How does this handle accessing region data structures in parallel??

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
```diff
@@ -205,10 +210,12 @@ public void onEnable() {
         }
         worldListener.registerEvents();
 
-        Bukkit.getScheduler().runTask(this, () -> {
+        getScheduler().runAsync(() -> {
```
Joo200, line 213: This runAsync call is not needed. You already create a task for every player in the for loop.

worldguard-bukkit/src/main/resources/plugin.yml
```diff
@@ -4,3 +4,4 @@ version: "${internalVersion}"
 depend: [WorldEdit]
 softdepend: [CommandBook]
 api-version: 1.19
+folia-supported: true
```
Joo200, line 7: missing newline

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/scheduler/SchedulerAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.sk89q.worldguard.bukkit.scheduler;
+
+import org.bukkit.Location;
+import org.bukkit.entity.Entity;
+
+public interface SchedulerAdapter {
```
Joo200, line 25: please add javadocs

Joo200 (commented):
I guess we can add Folia support at some point however there are some missing things:

1. WorldEdit should support Folia first. We don't support WorldEdit forks. Without Folia support for WorldEdit, WorldGuard runs in unsupported environment.
2. I'm not happy with the removed paper-api dependency. Our suggested platform is Paper, we should use paper as dependency. Folia adds some more methods and the bukkit module should never call them directly. This requires a new folia module.

> How does this handle accessing region data structures in parallel??

This should be fine, WorldGuard already supports multithreaded access. However we should double check that.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
```diff
@@ -990,8 +990,7 @@ public void onInventoryMoveItem(InventoryMoveItemEvent event) {
             handleInventoryHolderUse(event, cause, targetHolder);
 
             if (event.isCancelled() && causeHolder instanceof Hopper && wcfg.breakDeniedHoppers) {
```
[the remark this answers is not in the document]
HarvelsX: Does this piece of code even make sense if calling `event.isCancelled()` is always false when `ignoreCancelled = true`?

HarvelsX:
> 2\. I'm not happy with the removed paper-api dependency. Our suggested platform is Paper, we should use paper as dependency. Folia adds some more methods and the bukkit module should never call them directly. This requires a new folia module.

Creating the same code in every plugin is not reasonable. Maybe adopting this in [PaperLib](https://github.com/PaperMC/PaperLib/pull/66) would not be a bad thing.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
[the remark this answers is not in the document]
Joo200: It doesn't, you can remove the cancelled check.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
wizjany: this actually presents a race condition as running the init on region threads could actually happen after the async sessionmanager loop, leading to a player getting double-initialized.
that is, the bukkitsessionmanager itself should be handling scheduling for the individual player "events" (process/run)

Anon8281:
"Exit" and "Entry" flags do not work on Folia without throwing an exception. Perhaps this is caused by the use of Player#teleport, as Folia requires you to use teleportAsync

HarvelsX:
@Anon8281, it's kind of based on events/

Yomamaeatstoes:
Could you guys provide  temporary jar to test it?

Yomamaeatstoes:
I am sure you guys are aware of this, but someone told me the " it was one of the build.gradle.kts files " is incompatible in this PR. Once it is fixed it will be buildable. This was a test build that someone made for me, and I don't have access to make any commits here. However, with your permission I can post the jar here.

Loving11ish:
Hey folks, are you aware of the issue with one of the build.gradle.kts files that is currently preventing successful building?  Thanks in advance!

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
HarvelsX: @wizjany, creating a session does the initialization, so I don't see the need to call the initialization method in https://github.com/EngineHub/WorldGuard/blob/78498491ddf94f2fce376614e17556e5ff8f6327/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/session/BukkitSessionManager.java#L61 so can try to get a session to pre-create it.

HarvelsX:
@Yomamaeatstoes and @Loving11ish, fixed buggy rebase/

Yomamaeatstoes:
Hello @HarvelsX  I have just tested this, and it builds fine. Although I did find other issues. One of the issues is that it seems to disable NPC plugins like FancyNCP ( https://github.com/FancyMcPlugins/FancyNpcs ) , and ZNPC ( https://github.com/Pyrbu/ZNPCsPlus ) Although the NPC themselves spawn in, I have found that commands no longer work. When I uninstall WorldGuard they fully work again, and I have done a binary search. I am unaware what other issues if any have also been created. I did get this error log, so maybe this may help https://mclo.gs/Quv7LTJ

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitRegionContainer.java
```diff
@@ -94,7 +94,7 @@ public void onChunkUnload(ChunkUnloadEvent event) {
             }
         }, plugin);
 
-        Bukkit.getScheduler().scheduleSyncRepeatingTask(plugin, cache::invalidateAll, CACHE_INVALIDATION_INTERVAL, CACHE_INVALIDATION_INTERVAL);
+        plugin.getScheduler().runAsyncRate(cache::invalidateAll, CACHE_INVALIDATION_INTERVAL, CACHE_INVALIDATION_INTERVAL);
```
Joo200, line 97: This was sync before.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/session/BukkitSessionManager.java
```diff
@@ -60,8 +60,7 @@ public void resetAllStates() {
     @EventHandler
     public void onPlayerProcess(ProcessPlayerEvent event) {
         // Pre-load a session
-        LocalPlayer player = WorldGuardPlugin.inst().wrapPlayer(event.getPlayer());
-        get(player).initialize(player);
+        get(WorldGuardPlugin.inst().wrapPlayer(event.getPlayer()));
```
Joo200, line 63: why did you change that? it doesn't initialize the player anymore?

worldguard-bukkit/src/main/resources/plugin.yml
Joo200: missing newline *at end of the file*

Joo200 (commented):
The MR contains some improvements for Folia however without a WorldEdit version which supports Folia I wouldn't merge this.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/session/BukkitSessionManager.java
HarvelsX: @Joo200, initialization occurs during session creation, it is created when you try to get it from the cache. So there seems to be no need to specifically call the initialization method from here, instead you can try to create a session if it doesn't exist.

https://github.com/EngineHub/WorldGuard/blob/78498491ddf94f2fce376614e17556e5ff8f6327/worldguard-core/src/main/java/com/sk89q/worldguard/session/AbstractSessionManager.java#L178

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitRegionContainer.java
HarvelsX: @Joo200, if the initialization problem is solved, this can be left asynchronous.

Yomamaeatstoes:
> It works completely fine

I have since edited this message. This works completely fine.

HarvelsX:
@Anon8281, probably fixed, but there may be other problems.

Draghonir:
Hello,

I’m currently hosting a server based on Folia and using a community-made Folia-compatible build of WorldGuard, which is working without any issues so far.

That being said, I would personally feel more comfortable using an officially maintained version. WorldGuard plays a key role in server protection, and official Folia compatibility would provide greater confidence regarding long-term support, updates, and overall reliability.

I’m aware that Folia comes with important architectural differences compared to Paper and that adding support may not be trivial. I mainly wanted to share my interest and support for official Folia compatibility for WorldGuard, should this be something you plan to explore in the future.

Thank you for your time and for continuing to maintain such an essential plugin for the Minecraft server community.

Issue #2013, opened 2023-06-09
/locate command from worldguard overwrites the vanilla one


CryRocks:
### The Problem

the /locate command from worldguard, overwrites the /locate biome/structure command from the vanilla game.

### A Solution

maybe add the option to the worldguard config, to activate/deactivate that command or change the command name so that /locate is not overwritten and the vanilla function can be used.

### Anything Else?

maybe i'm just blind and didn't find the option in the config or so but i checked and couldn't find any.
Also setting the permission from the command to false, didn't work.

me4502:
This is up to the platform to handle, not plugins. Paper provides a command.yml file to handle this, WorldGuard cannot change what overwrites what

CryRocks:
> This is up to the platform to handle, not plugins. Paper provides a command.yml file to handle this, WorldGuard cannot change what overwrites what

oh i see, tried a bit of stuff but it says unknown command:

```
  locate:
  - minecraft:locate
  wglocate:
  - worldguard:locate
```

second one works, but something is wrong with the first one, do you maybe know how it done correctly? would appreciate the help :)

me4502:
You'd have to ask Paper for help

xpdlx:
I'm having the same problem with WorldGuard overriding /heal in my plugin.

Having a way to disable certain WorldGuard commands would be nice. 
I tried overriding it in commands.yml, which worked, but that doesn't seem to support command completions, so that's pretty annoying.

I also tried negating the permission for it, but that just made it say "You don't have permission to do that." instead of using my command.

I fixed it by making my plugin load before WorldGuard, but it would be nice if there was a way in WorldGuard to do this (either by negating the permission, or have a thing in the config to disable commands).

black-kirito:
you can use the minecraft: tag example: /minecraft:locate biome minecraft:savanna

Leahcimkrob:
Why should you enter a standard command in such a complicated way just because you overwrote it? We don't need your locate command, but we have minecraft command loacte in rank system. Please make the worldguard:locate command deactivatable, or use another one.

me4502:
The WorldGuard locate command existed for a very very long time before the Minecraft one. If your server platform has decided it’s taking precedence, you should use the systems that they provide you to let the Minecraft command take precedence. If there are issues with that system that should be a bug report to your server platform.

Leahcimkrob:
under MC 1.16 we already had that. and now with the new Paper (1.20) no more.

[...]

PR #2041, opened 2023-11-07
Add command plugin compatability into GeneralCommands

codeHusky:
CommandBook compat within WorldGuard was gutted some time ago, but the GeneralCommands don't register when CommandBook is present. This functionality is not identical for similar plugins such as Essentials, which have their commands unexpectedly overridden by WorldGuard.

It's worth noting that all functionality within GeneralCommands is not required for the continued functionality of the plugin. CommandBook support for the CommandBook god mode and detecting it is **entirely disabled** in the current version of WorldGuard, only existing in references to the plugin in the documentation and commented-out code. If this functionality is desired, it should be implemented in a more relevant PR or elsewhere.

This PR makes this functionally identical when using the similar-featured and popular plugin Essentials. Functionality could be added to support other Essentials-like plugins via changing the wording to "Third Party" rather than Essentials and some sort of list, but most complaints seem to be from folks using Essentials.

This PR also adds "GeneralCompatibleCommands", which is basically a wrapper for GeneralCommands with different aliases. I'm not personally familiar with the WorldEdit command system and it seemed the "cleanest" way to handle the alias removal situation would be to just have the wrapper class.

Fixes issue #2013

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitConfigurationManager.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/BukkitConfigurationManager.java
@@ -37,6 +37,7 @@ public class BukkitConfigurationManager extends YamlConfigurationManager {
 
     private boolean hasCommandBookGodMode;
     boolean extraStats;
+    private boolean hasEssentialsGodMode;
 
     /**
      * Construct the object.
@@ -123,5 +124,13 @@ public class BukkitConfigurationManager extends YamlConfigurationManager {
 
     public boolean hasCommandBookGodMode() {
         return hasCommandBookGodMode;
+    }
+
+    public void updateEssentialsGodMode() {
+        hasEssentialsGodMode = plugin.getServer().getPluginManager().isPluginEnabled("Essentials");
+    }
+
+    public boolean hasEssentialsGodMode() {
+        return hasEssentialsGodMode;
     }
 }
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
@@ -160,7 +160,10 @@ public class WorldGuardPlugin extends JavaPlugin {
         reg.register(ToggleCommands.class);
         reg.register(ProtectionCommands.class);
 
-        if (!platform.getGlobalStateManager().hasCommandBookGodMode()) {
+        platform.getGlobalStateManager().updateCommandBookGodMode();
+        platform.getGlobalStateManager().updateEssentialsGodMode();
+
+        if (!platform.getGlobalStateManager().hasCommandBookGodMode() && !platform.getGlobalStateManager().hasEssentialsGodMode()) {
             reg.register(GeneralCommands.class);
         }
 
@@ -191,8 +194,6 @@ public class WorldGuardPlugin extends JavaPlugin {
         if ("true".equalsIgnoreCase(System.getProperty("worldguard.debug.listener"))) {
             (new DebuggingListener(this, WorldGuard.logger)).registerEvents();
         }
-
-        platform.getGlobalStateManager().updateCommandBookGodMode();
 
         if (getServer().getPluginManager().isPluginEnabled("CommandBook")) {
             getServer().getPluginManager().registerEvents(new WorldGuardCommandBookListener(this), this);
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/WorldGuardPlugin.java
@@ -63,6 +63,7 @@ import com.sk89q.worldguard.bukkit.util.ClassSourceValidator;
 import com.sk89q.worldguard.bukkit.util.Entities;
 import com.sk89q.worldguard.bukkit.util.Events;
 import com.sk89q.worldguard.commands.GeneralCommands;
+import com.sk89q.worldguard.commands.GeneralCompatibleCommands;
 import com.sk89q.worldguard.commands.ProtectionCommands;
 import com.sk89q.worldguard.commands.ToggleCommands;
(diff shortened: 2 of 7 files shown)


ThePotatoDev:
@wizjany 🙏

Joo200:
I don't think it's a good idea to add Essentials as a new soft-depend. Maybe there is a better way to detect registered commands in existing plugin.yml from other plugins.

At some point we should remove the soft-depend to CommandBook and handle the commands only from worldguard. The command aliases prefixed by /wg are fine but the implementation is terrible.

The paper logic in SimpleCommandMap has the following comment when registering a command:
`// If the command exists but is an alias we overwrite it, otherwise we return`
We should be able to use that by using the /wg-prefixed command labels as first command in the annotation.

codeHusky:
@Joo200 I attempted this but it doesn't appear Essentials overrides the commands in that situation. The commands are still handled by WorldGuard
```java
 @Command(aliases = {"wggod", "god"}, usage = "[player]",
            desc = "Enable godmode on a player", flags = "s", max = 1)
    public void god(CommandContext args, Actor sender) throws CommandException, AuthorizationException {
        Iterable<? extends LocalPlayer> targets = null;
```
Essentials is trying to not override other plugins with its command registrations deliberately, as most of the time that's what people want. Some changes could be made to how these commands are created in the first place (using the non-annotation way of defining commands from WorldEdit) but we still need to ensure WorldGuard loads last so we can, at minimum, check the command map to see if "heal", etc are registered

codeHusky:
Realistically then, these commands should be disabled by default. They override other plugins commands and provide functionality that isn't core and essential to WorldGuard's function. Keep them enabled for existing installs, but new configs should probably start with them disabled. As I mention in #2042, "[Everything is off by default](https://github.com/EngineHub/WorldGuard/blob/master/README.md)" is one of the key features of WorldGuard and this behavior is rather contradictory

wizjany:
yea this is somewhat philosophically misguided from the onset. the reason we tested specifically for commandbook is because we *integrated* with it. testing for the presence of essentials is useless - if we just want to avoid command conflicts, we need to be testing if the command is registered, *not* if some *specific plugin* is loaded (never mind enabled and registering the command itself). an essentials version of the GodMode session handler can be added easily by a third party plugin, but we shouldn't pretend to depend on essentials if we're not actually using their api to do things properly (and i'd like to avoid third party apis anyway).

also you're entirely mis-reading "off by default". that refers to changing default game behavior. using a command to turn on god mode *is opting in to changed behavior*. this is a nonsense argument.

re: onCommand, no that's very much part of the api, and it still requires command registration. plugin#onCommand is called for the owner(plugin) of the command *if it doesn't have an executor set*. i.e. it's a fallback/catchall for plugins that don't use individual executors, such as those using their own handling logic (e.g. most command frameworks)

codeHusky:
Given how many issues have been made about these commands being on by default being a "bug" of some kind, I can't help but think that the interpretation of "off by default" you have conflicts with what people are taking it as. I don't see a point in having these commands even strictly included on by default with WorldGuard is in the first place other than "it's always been there." They don't serve most users needs very well in the first place, and the "locate" command especially replaces something very useful for SMP admins with something intrusive and unhelpful.

If long-time users want to use them, they should be able to use them, but many of them override high-use existing features due to their current aliases. The bulk of Bukkit/Spigot/Paper servers use Essentials, and many modern servers want to use vanilla /locate, so it seems a bit pointless to try and keep these commands for new installs. It's not like a fresh config having them off by default will exactly break compat with anything either.

Essentials itself goes the correct route in this situation - commands that get replaced by other plugins have a fallback `/e<cmd>` alias for each. Yes, the `/wg<cmd>` alias is ugly but the current functionality is rather gross as-is. I'd love it if the Bukkit namespacing was a perfect solution here but it's not, and choosing to simply not support one of the most popular plugins on Bukkit servers because it's not something you feel like you should have to deal with is a bit silly. 

If you don't want a softdep then I guess configuring the time the plugin tries to load itself in the plugin.yml to be late vs essentials and other plugins would make the most sense. Regardless, not having any sort of compatibility measure just seems pointlessly aggressive for a plugin that's primary function is not to provide helpful commands to users. Players hate that kind of stuff - same reason many people don't use certain "Anti Chat Reporting" plugins that replace the messaging commands, etc. It's unexpected features that interfere with things they have deliberately added to their server for a certain purpose. Let's not make server admin more of a headache for people than necessary - it's hard enough to figure out what plugin is providing these pink-text god/ungod commands in the first place.

Joo200:
> I attempted this but it doesn't appear Essentials overrides the commands in that situation.

This is configured by the Essentials config. By adding some prefix to the worldguard gods command and registering the command "god" as alias I was able to configure Essentials with the overridden-commands config to use the essentials god command. Imo this should be the preferred solution for now. Unfortunatly it's not possible to hide them by default as a /wg subcommand.

Indeed the onCommand method is part of Bukkti API, I didn't see Essentials has some workaround for other plugins.

codeHusky:
I'd still, personally, prefer a config option to just nuke the commands if this particular PR isn't merged. I'd rather not have to fudge with the massive Essentials config to fix an issue that's ultimately on WG's side, especially with commands I'll never use.

Joo200:
I'm not gonna argue about which plugin is reasonable for this issue. It's not only WG's side. If you don't want to modify Essentials config that's your problem.

> Realistically then, these commands should be disabled by default. They override other plugins commands and provide functionality that isn't core and essential to WorldGuard's function. Keep them enabled for existing installs, but new configs should probably start with them disabled.
It's not possible to create a new config value which is disabled by default but enabled for existing installations. I don't mind a new config option to disable those commands but the commands should be enabled by default.

codeHusky:
> It's not possible to create a new config value which is disabled by default but enabled for existing installations.
I mean, that's a bit of an exaggeration - see [this commit](https://github.com/EngineHub/WorldGuard/pull/2042/commits/7c7edc7d5fe74d2705918a12868cdb5f09a8e152). Works perfectly in my testing.

wizjany:
we will not be breaking default behavior.

codeHusky:
I'm not sure what you mean, the commit does not break any default behavior. It's a new behavior that disables legacy features by default to fit more in line with what players seem to expect. If you mean that you don't want to disable the commands by default, I'm curious for reasoning as to why. Are you aware of them being used a lot on servers?

PR #2082, opened 2024-05-18
Fix checking bypass on initialize

aromaa:
Fixed that checking bypass on handlers initialize would always return false. Not really sure if this is the best approach.

Also removed the `SessionManager#createSession` because it was making this a bit more complicated and its not really used.


wizjany:
Removing `createSession` is technically an API-break even though that shouldn't realistically be getting called from outside WG. Can we maybe just deprecate it for removal and have it return `get()` to maintain some semblance of just-in-case compatibility?

PR #2083, opened 2024-05-24
Adds concrete powder/anvil to no-physics-config

### LINKED ISSUE #1411: Add more blocks to physics control
WORLDGUARD-3884 - Reported by CoolDev457

Hello,

In the physics config :
physics:
    no-physics-gravel: true
    no-physics-sand: true
	
It would be nice, with the new blocks of 1.11, to add no-physics-concrete-powder: true
And maybe at the same time no-physics-anvil.

Best regards,

MFreeware

aurorasmiles:
Fixes #1411
This feels like something often requested in Discord. There are in theory a couple more blocks that can be affected by gravity, but most of those seem like edge cases that are rarely relevant (except the dragon egg maybe, but I feel like the dragon egg has a lot more annoying things to block before gravity)


wizjany:
I've been mentally pushing this off every time someone asks because I feel like adding a million config options isn't the way to go long-term. Perhaps it's better to have a list like the water/lava/fire spread configs?

InstantlyMoist:
This seems fine for now no? Better than nothing and having to make/install another plugin just for this.

PR #2085, opened 2024-06-04
Special handling for Windcharges

Brokkonaut:
WindCharges cause explosions but they never break blocks. So I removed the call of BlockBreakEvent for them and call UseBlockEvents instead, because wind charges may "use" blocks and added a separate flag.

I tried to use the UseBlockEvent as bulk event but then it will always check the same material for all locations, so that did not work. Thats why I use separate events for every block

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
@@ -65,6 +65,7 @@ import org.bukkit.block.PistonMoveReaction;
 import org.bukkit.block.data.BlockData;
 import org.bukkit.block.data.Waterlogged;
 import org.bukkit.block.data.type.Dispenser;
+import org.bukkit.entity.AbstractWindCharge;
 import org.bukkit.entity.AreaEffectCloud;
 import org.bukkit.entity.Creeper;
 import org.bukkit.entity.Entity;
@@ -79,7 +80,6 @@ import org.bukkit.entity.Painting;
 import org.bukkit.entity.Player;
 import org.bukkit.entity.Tameable;
 import org.bukkit.entity.ThrownPotion;
-import org.bukkit.entity.WindCharge;
 import org.bukkit.entity.minecart.HopperMinecart;
 import org.bukkit.event.Cancellable;
 import org.bukkit.event.Event;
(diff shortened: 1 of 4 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
```diff
@@ -357,6 +359,19 @@ public void onEntityChangeBlock(EntityChangeBlockEvent event) {
     @EventHandler(ignoreCancelled = true)
     public void onEntityExplode(EntityExplodeEvent event) {
         Entity entity = event.getEntity();
+        if (entity instanceof WindCharge) {
```
Joo200, line 362: I'm not happy with specifying WindCharge at this location. Explosions can have different interaction types:
* KEEP: Not changing anything
* DESTROY, DESTROY_WITH_DELAY: Destroy blocks
* TRIGGER_BLOCK: trigger the blocks.

The current changes check only for WindCharge but not for BreezeWindCharge which also triggers blocks but don't destroy them.

Imo the correct approach should be that Spigot (or Paper) adds the Interaction type of a explosion to EntityExplodeEvent and BlockExplodeEvent. After that we can change our implementation to handle the interaction correctly.

Brokkonaut: > Imo the correct approach should be that Spigot (or Paper) adds the Interaction type of a explosion to EntityExplodeEvent and BlockExplodeEvent. After that we can change our implementation to handle the interaction correctly.

I agree that Spigot or Paper would have to add api to correctly fix this, that check is just a workaround until that api is added. For the specific issue, I could change the check to AbstractWindCharge in the pr, I just inadvertently thought that BreezeWindCharge is a subclass of WindCharge.

Joo200: Created https://hub.spigotmc.org/jira/browse/SPIGOT-7799

Joo200:
I changed some more code of your MR and added it directly to the mc-1.21 feature branch.

We have some inconsistency in WorldGuard for the internal events when multiple blocks are affected. I changed that so we can handle the explosion with a single UseBlockEvent.

Thanks for the contribution.

PR #2089, opened 2024-06-17
Add support for MC 1.21.

wizjany:
* [ ] ~~Change material flag map from enummap to non-enum map. (probably doesn't need more than this, depending on if/when this happens, we may just want to compile on the new API directly)~~
* [x] See what other stuff got added in MC 1.21, #2085 
* [x] idk

~~nothing has been tested yet~~

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/BlacklistListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/BlacklistListener.java
@@ -96,7 +96,7 @@ public class BlacklistListener extends AbstractListener {
         event.filter(target -> {
             if (!wcfg.getBlacklist().check(
                     new BlockBreakBlacklistEvent(localPlayer, BukkitAdapter.asBlockVector(target),
-                            createTarget(target.getBlock(), event.getEffectiveMaterial())), false, false)) {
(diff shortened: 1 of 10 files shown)


dead-plant:
can u please give me a .jar file?

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/BlacklistListener.java
```diff
@@ -96,7 +96,7 @@ public void onBreakBlock(final BreakBlockEvent event) {
         event.filter(target -> {
             if (!wcfg.getBlacklist().check(
                     new BlockBreakBlacklistEvent(localPlayer, BukkitAdapter.asBlockVector(target),
-                            createTarget(target.getBlock(), event.getEffectiveMaterial())), false, false)) {
+                            createTarget(target.getBlock())), false, false)) {
```
Joo200, line 99: Instead of using the effectiveMaterial I changed this to the actual block for on-break, on-place and on-interact. This has some implications:

* Now the actual block is checked. If someone is destroying the world with TNT the destroyed block type is checked and not the TNT material.
* on-use can be used to check to deny usage of stuff. e.g. on-use[WIND_CHARGE] will block the usage of wind charges, on-interact[OAK_DOOR] will block the interaction of the wind charge with the door.

PR #2093, opened 2024-06-19
Fix ItemStack conversion on block place on MC 1.21

lenis0012:
This issue is present in 1.21 because some materials do not convert 1:1 to item types.  
This throws an exception in the ItemStack constructor.  
Example: WALL_TORCH is a valid block but not a valid item, only TORCH is a valid item.

So, use the item in hand instead, creating a copy with quantity set to 1, for issuing the use item event.

--------

To reproduce: place a torch on the wall.

If a different solution is preferred that is fine too and this MR may be closed but I don't see an issue with using the item in hand as I believe it is only changed after the event has completed, so it will contain the item placed even if the quantity was 1.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
@@ -218,8 +218,8 @@ public class EventAbstractionListener extends AbstractListener {
             Events.fireToCancel(event, new BreakBlockEvent(event, create(event.getPlayer()), previousState.getLocation(), previousState.getType()));
         }
 
-        if (!event.isCancelled()) {
-            ItemStack itemStack = new ItemStack(event.getBlockPlaced().getType(), 1);
+        if (!event.isCancelled() && !event.getItemInHand().isEmpty()) {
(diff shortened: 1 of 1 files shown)


lenis0012:
Related to #2089

wizjany:
the quantity part shouldn't even be necessary, we only care about the item type to check the interaction mappings. and according to the javadocs, that should return the itemstack at the time of placement, so it shouldn't be consumed until after the event either.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
```diff
@@ -218,8 +218,8 @@ public void onBlockPlace(BlockPlaceEvent event) {
             Events.fireToCancel(event, new BreakBlockEvent(event, create(event.getPlayer()), previousState.getLocation(), previousState.getType()));
         }
 
-        if (!event.isCancelled()) {
-            ItemStack itemStack = new ItemStack(event.getBlockPlaced().getType(), 1);
+        if (!event.isCancelled() && !event.getItemInHand().isEmpty()) {
+            ItemStack itemStack = event.getItemInHand().asQuantity(1);
```
wizjany, line 222:
```suggestion
            ItemStack itemStack = event.getItemInHand();
```

PR #2100, opened 2024-07-03
Port buildlogic from WorldEdit

aurorasmiles:
This tries to port the current build logic from WorldEdit to WorldGuard.
The project compiles fine, and the jars look similar enough from a quick glance. Limited testing in Minecraft showed no apparent issues. The only differences I've noticed:
- pom of prtree/flyway-core no longer in META-INF
- some bstats classes no longer included
- some flyway classes no longer included

Notes:
- Piston stuff is currently commented out, but I decided to still leave it in - should make transitioning easier at some point 
- I've commented out `-Werror` in the compilerArgs for now - ~~suppressing~~ addressing all of the deprecation warnings should be a seperate PR
- Something else that still needs to be ported are the checkstyle changes, but again, not here
- There *might* (read: probably will) be some WorldEdit stuff leftover that isn't required in WorldGuard, even though I went over everything multiple times


gradle/libs.versions.toml
```diff
@@ -0,0 +1,86 @@
...
+paperweight = "io.papermc.paperweight.userdev:io.papermc.paperweight.userdev.gradle.plugin:1.7.1"
+
+squirrelid = "org.enginehub:squirrelid:0.3.2"
+
+autoService = "com.google.auto.service:auto-service:1.1.1"
+
+jsr305 = "com.google.code.findbugs:jsr305:3.0.2"
+jetbrains-annotations = "org.jetbrains:annotations:24.1.0"
+#errorprone-annotations = "com.google.errorprone:error_prone_annotations:2.11.0"
+#
```
octylFractal, line 31:
```suggestion
```

octylFractal (approved):
LGTM if you fix the end of file newlines (they should be present)

PR #2110, opened 2024-07-14
Update WorldGuardEntityListener.java

Lory9098:
add this to fix some mods and plugin problem


wizjany:
hybrids aren't supported in the first place, and all those platforms have to do is add proper mappings between the MC registry and the bukkit enum. don't want to add hacky workarounds on our end for this.

PR #2150, opened 2024-10-16
Avoid using metadata storage if possible

### LINKED ISSUE #1535: Shaky-sand caused memory leak
WORLDGUARD-4169 - Reported by djytw

(Version: WorldGuard-7.0.2 on 1.14.4)

My server always run out of memory after several hours, and starts full GC, with no effect. Memory usage is still high after lots of Full  GC, and server will lag with tps under 5.

After some test, I found this is caused by the 0tick shaky sand and WorldGuard. To validate this, I created a new map with ~300 shaky sand, using a common method (boat on a campfire). With WorldGuard installed, the server will run out 3G of memory and tps drop to 5 in half an hour; on the other hand, without the plugin, and using the same map, the server runs without problem for days.

I use jmap and MAT to analyse the heap, found this line: https://github.com/EngineHub/WorldGuard/blob/6f6125fe00022b4b1167c12ac5bdeebf48318d5b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java#L236 occupies most of the memory, which might be called by a EntityChangeBlockEvent:https://github.com/EngineHub/WorldGuard/blob/6f6125fe00022b4b1167c12ac5bdeebf48318d5b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java#L292 . 

I think every time the boat changes the sand between entity and block, these code will add a tracker information to it, then occupy a huge amount of memory.

This is quite a crutial problem since there's an increasing number of players start using this technique to build some farming mechine, and I have to restart the server 4 times a day to free the memory, or it will be extremely lag. Although this method of creating shaky-sand is fixed in 1.15, there're many new methods can also do the same. Futhermore, villager farming and ender dragon and wither destroying blocks might also have the same issue, since the problem is introduced by the EntityChangeBlockEvent.

Joo200:
Currently the sources for creeper explosions and falling blocks are stored in the metadata section of bukkit. The metadata section is never cleared so with a lot of falling blocks there is a memory leak.

This PR removes the tracking for Creeper explosions since there is already `Creeper#getIgniter()`.

For Paper we use `FallingBlock#getOrigin()` ~~if the config options usePaperEntityOrigin is enabled.~~

Closes #1535

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
@@ -76,7 +76,7 @@ public final class Cause {
     /**
      * Create a new instance.
      *
-     * @param causes   a list of causes
+     * @param causes a list of causes
      * @param indirect whether the cause is indirect
      */
     private Cause(List<Object> causes, boolean indirect) {
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
@@ -21,7 +21,6 @@ package com.sk89q.worldguard.bukkit.cause;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.Sets;
-import com.sk89q.worldedit.bukkit.BukkitAdapter;
 import com.sk89q.worldguard.bukkit.BukkitWorldConfiguration;
 import com.sk89q.worldguard.bukkit.WorldGuardPlugin;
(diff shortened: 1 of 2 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
```diff
@@ -74,7 +76,7 @@ public final class Cause {
     /**
      * Create a new instance.
      *
-     * @param causes a list of causes
+     * @param causes   a list of causes
```
aurorasmiles:
```suggestion
     * @param causes a list of causes
```

aurorasmiles (approved):

PR #2180, opened 2025-01-27
add ability to disable region in region

jankornfeld:
Adds the ability to prevent users from creating a region which is within another region.
This is configurable and is per default set to false (which is the normal behavior)

I added this because I needed to prevent my players or trolls to protect over another players region.
Dunno if this is already possible but this here is working fine for me.

image


wizjany:
Maybe you should be using claim instead of define? Define is very specifically meant to not have the restrictions that claim does.

PR #2193, opened 2025-03-12
Listen to EntityPushedByEntityAttackEvent for knockback/push

### LINKED ISSUE #2190: Mace smash attack AoE knockback not handled
### WorldEdit Version

7.3.10

### WorldGuard Version

7.0.13

### Platform Version

Paper version 1.21.4-204-main@3d13b11

### Confirmations

### Bug Description

Hitting a mob with a mace and causing a smash attack (falling from >1.5 blocks before hitting) in a region that has `pvp: deny` causes an AoE knockback effect that knocks other entities as well. This results in the ability of players to knock others out of a region by indirectly hitting other mobs with this attack. (assuming animal / mob harm is allowed to land the attack)

Even with animal harm denied it's still possible to do this with mobs.

### Expected Behavior

The indirect knockback resulting from the smash attack AoE being denied.

### Reproduction Steps

1. Set up a region with flags: `pvp: deny`
2. Have a player and another mob standing close to each other
3. Jump from >1.5 blocks and hit the mob with a mace causing a smash attack

**Result** the other player gets sent flying away

_No response_

### Anything Else?

Fixing this issue most likely requires Paper and Spigot calling `EntityKnockbackByEntityEvent` on other entities affected by the smash attack AoE. This does not yet happen, however I've opened an issue with both to see if that were possible.

_If I understand correctly since Paper hardforked they are not pulling changes from Spigot, hence two reports._ \
[Paper issue](https://github.com/PaperMC/Paper/issues/12253)
[Spigot issue](https://hub.spigotmc.org/jira/browse/SPIGOT-8016)

I am not sure what changes would be required in WG for this to be protected properly. My assumptions would be either include this behaviour under the pvp flag or create a flag specifically for knockback.

There is also [`EntityAttemptSmashAttackEvent`](https://jd.papermc.io/paper/1.21.4/io/papermc/paper/event/entity/EntityAttemptSmashAttackEvent.html) available on Paper that could be used to deny using smash attacks specifically. Adding a flag for that is also what I did in order to overcome the issue. (and tried submitting a [PR](https://github.com/aromaa/WorldGuardExtraFlags/pull/282) to WG extra flags)

Relevant discussion on the discord server: https://discordapp.com/channels/80853743897149440/470410381597343753/1282704841226850365

Wertik:
Listening to the push event instead of EntityKnockbackByEntity includes other sources of knockback/push in the internal DamageEntityEvent.

For spigot, after [this PR](https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/pull-requests/1538/overview) is merged, WG should handle mace AoE knockback as it is.

This is primarily to include the Mace AoE knockback in the pvp flag, but affects other sources of push/knockback as well. Essentially all calls of Entity#push (ex.: ender dragon pushes, arrows, creaking, ravagers). I don't think this should cause any unexpected behaviour, but another opinion would be great.

I tested the effect of this on the pvp flag. It does prevent the knockback/push of the Mace AoE when pvp is denied, but does send a "cannot pvp here" message. Not sure if that's wanted here as it might be confusing since the player is not attacking anyone directly.

Closes #2190

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
@@ -47,6 +47,7 @@ import com.sk89q.worldguard.bukkit.util.Materials;
 import com.sk89q.worldguard.config.WorldConfiguration;
 import com.sk89q.worldguard.protection.flags.Flags;
 import io.papermc.lib.PaperLib;
+import io.papermc.paper.event.entity.EntityPushedByEntityAttackEvent;
 import io.papermc.paper.event.player.PlayerOpenSignEvent;
 import org.bukkit.Bukkit;
 import org.bukkit.Effect;
@@ -1324,8 +1325,8 @@ public class EventAbstractionListener extends AbstractListener {
         }
 
         @EventHandler(ignoreCancelled = true)
-        public void onEntityKnockbackByEntity(com.destroystokyo.paper.event.entity.EntityKnockbackByEntityEvent event) {
-            handleKnockback(event, event.getHitBy());
+        public void onEntityKnockbackByEntity(EntityPushedByEntityAttackEvent event) {
+            handleKnockback(event, event.getPushedBy());
         }
     }
 


wizjany:
As far as I can tell from [your comment](https://github.com/PaperMC/Paper/pull/12257#issuecomment-2711817183), the EKBE handler will fire on an EPBEAE as well. Why do we need to change the event here? It seems that EKBE should be a superset of EPBEAE, not a subset?

edit: ok had to pull up the javadocs. I don't understand the naming ig but the hierarchy is
```
io.papermc.paper.event.entity.EntityKnockbackEvent
    io.papermc.paper.event.entity.EntityPushedByEntityAttackEvent
        com.destroystokyo.paper.event.entity.EntityKnockbackByEntityEvent
```

As far as the PVP messages, if that's something that gets spammy, the internal DamageEntityEvent can be setSilent in the common knockback handler. I don't think it makes sense to make it indirect as the knockback pretty directly comes from the player's attack imo.

Wertik:
> As far as I can tell from [your comment](https://github.com/PaperMC/Paper/pull/12257#issuecomment-2711817183), the EKBE handler will fire on an EPBEAE as well. Why do we need to change the event here? It seems that EKBE should be a superset of EPBEAE, not a subset?
> 
> edit: ok had to pull up the javadocs. I don't understand the naming ig but the hierarchy is
> ```
> io.papermc.paper.event.entity.EntityKnockbackEvent
>     io.papermc.paper.event.entity.EntityPushedByEntityAttackEvent
>         com.destroystokyo.paper.event.entity.EntityKnockbackByEntityEvent
> ```
> 
> As far as the PVP messages, if that's something that gets spammy, the internal DamageEntityEvent can be setSilent in the common knockback handler. I don't think it makes sense to make it indirect as the knockback pretty directly comes from the player's attack imo.

I can see how my comments and code there might be confusing. I had a listener for EKBE to catch them all. Could've been EPBEAE instead.

The naming of the events doesn't help, I had troubles distinguishing when they should get called as well. There's no explanation in the javadocs about the difference between EKBBEE (EntityByEntity) and EPBEAE.

If I got it right you agreed with the choice after looking at the hierarchy?

It might get spammy when there are more players around (message for each player), but since that rarely happens I think we're ok. My main concern was the possible confusion, it's a minor thing though and if you lean that way that's all I need, let's keep it verbose.

edit: btw. Spigot PR got merged, I'll test tomorrow (GMT+1), but it should work as it is

Wertik:
Adding a log to the SpigotListener for EKBEE....
```
[09:58:13] [Server thread/INFO]: Target: SHEEP Source: PLAYER
[09:58:13] [Server thread/INFO]: Target: PLAYER Source: SHEEP
```
Doesn't look right, time to reopen on spigot

Wertik:
Fixed on spigot's end, was being called with the wrong entity as attacker. If you agree with the event change for paper it should be good to go.

PR #2204, opened 2025-04-11
Update to 1.21.5.

wizjany:
It doesn't seem like there's much new behavior in this game drop. In theory, the existing release should work through commodore with the exception of overprotecting these new blocks.


F3hliks:
Are the messages clickable?

wizjany:
whatever your issue is likely has nothing to do with worldguard and definitely has nothing to do with this branch.

F3hliks:
> whatever your issue is likely has nothing to do with worldguard and definitely has nothing to do with this branch.

I've heard someone else say that clickable messages had to be re-done for 1.21.5 plugins but I haven't tried anything yet

Issue #2216, opened 2025-06-30
[Paper 1.21.7] Legacy PlayerLogihnEvent Detected


x9nico:
### WorldEdit Version

FastAsyncWorldEdit-Paper-2.13.1-SNAPSHOT.jar (but it's not linked of this issue)

### WorldGuard Version

bukkit-7.0.13

### Platform Version

Paper version 1.21.7-8-main@6e021c8 (2025-06-30T18:06:02Z) (Implementing API version 1.21.7-R0.1-SNAPSHOT)

### Confirmations

- [x] I am using the most recent Minecraft release.
- [x] I am using a version of WorldEdit compatible with my Minecraft version.
- [x] I am using a version of WorldGuard compatible with my Minecraft version.
- [x] I am using the latest or recommended version of my platform software.
- [x] I am NOT using a hybrid server, e.g. a server that combines Bukkit and Forge. Examples include Arclight, Mohist, and Cardboard.
- [x] I am NOT using a fork of WorldEdit, such as FastAsyncWorldEdit (FAWE) or AsyncWorldEdit (AWE)

### Bug Description

When I'm joining my Paper 1.21.7 Server, it says WorldGuard use deprecated `PlayerLoginEvent`.

![Image](https://github.com/user-attachments/assets/b9df8da2-1760-4ff9-8d25-6b6dc9cd252b)

### Expected Behavior

It should works

### Reproduction Steps

1. Create Paper 1.21.7 Server
2. Upload WorldGuard into your Server
3. Join your Server

### Optional WorldGuard-Report

https://paste.enginehub.org/1Ngc_zC-j.report

x9nico:
Any news about this ??

me4502:
There's still no alternative available in Paper for us to move away from
this event, so there's nothing we can do until that's possible

PR #2219, opened 2025-07-19
Update deprecated PlayerLoginEvent to AsyncPlayerPreLoginEvent

Woodstop:
This PR updates the plugin to replace usage of the deprecated `PlayerLoginEvent` event with `AsyncPlayerPreLoginEvent` per issue #2216, following the latest Bukkit API deprecations.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardPlayerListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardPlayerListener.java
@@ -49,15 +49,7 @@ import org.bukkit.event.Event;
 import org.bukkit.event.EventHandler;
 import org.bukkit.event.EventPriority;
 import org.bukkit.event.block.Action;
-import org.bukkit.event.player.AsyncPlayerChatEvent;
-import org.bukkit.event.player.PlayerCommandPreprocessEvent;
-import org.bukkit.event.player.PlayerGameModeChangeEvent;
-import org.bukkit.event.player.PlayerInteractEvent;
-import org.bukkit.event.player.PlayerItemHeldEvent;
-import org.bukkit.event.player.PlayerJoinEvent;
-import org.bukkit.event.player.PlayerLoginEvent;
-import org.bukkit.event.player.PlayerRespawnEvent;
-import org.bukkit.event.player.PlayerTeleportEvent;
+import org.bukkit.event.player.*;
 import org.bukkit.event.player.PlayerTeleportEvent.TeleportCause;
 import org.bukkit.inventory.ItemStack;
 
@@ -66,6 +58,8 @@ import java.util.Set;
 import java.util.logging.Logger;
 import java.util.regex.Pattern;
 import javax.annotation.Nullable;
+
+import static org.bukkit.Bukkit.getPlayer;
 
 /**
  * Handles all events thrown in relation to a player.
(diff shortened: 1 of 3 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardPlayerListener.java
```diff
@@ -168,13 +167,14 @@ public void onPlayerChat(AsyncPlayerChatEvent event) {
     }
 
     @EventHandler(ignoreCancelled = true)
-    public void onPlayerLogin(PlayerLoginEvent event) {
-        Player player = event.getPlayer();
+    public void onPlayerLogin(AsyncPlayerPreLoginEvent event) {
```
me4502, line 170: Does this retain Spigot compatibility? From memory, that was the early concern with moving to this event

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardPlayerListener.java
```diff
@@ -134,6 +129,10 @@ public void onPlayerJoin(PlayerJoinEvent event) {
         Events.fire(new ProcessPlayerEvent(player));
         WorldGuard.getInstance().getExecutorService().submit(() ->
             WorldGuard.getInstance().getProfileCache().put(new Profile(player.getUniqueId(), player.getName())));
+
+        if (cfg.deopOnJoin) {
+            player.setOp(false);
```
me4502, line 134: Is it possible for the player to perform any actions (eg, send commands) prior to this event being fired?

From memory, this event is fired once _everything_ was loaded, and at least in the past the player had partial capabilities prior to this point. Not sure if that's been resolved with the various new stages in the protocol during login though.

If that's the case, it might actually remove some of the purpose of this config option

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardPlayerListener.java
Woodstop: Yes, this is compatible with Spigot API. It has been stable for many versions and is still present up to version 1.21.8. This event is safe to use for async pre-login logic that doesn't require access to the Player object.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardPlayerListener.java
Woodstop: I think that's correct. Moving the deopOnJoin logic to PlayerJoinEvent *could* introduce a small window in which an OP player could technically execute a privileged command before being deopped. Since AsyncPlayerPreLoginEvent doesn’t provide a Player object, there might not be a clean way to deop earlier without relying on deprecated methods.

Here's a possible workaround that will check if the player is OP when the config is set to de-op on join before they send a command. The issue is that if the player meets both requirements (is both an OP and should be de-opped on join), it will prevent them from sending any commands at all until everything has fully loaded. Because that is a small or even negligible time window, this could be acceptable. I can't think of another way at the moment. 

```
    @EventHandler(priority = EventPriority.HIGHEST)
    public void onCommand(PlayerCommandPreprocessEvent event) {
        Player player = event.getPlayer();
        ConfigurationManager cfg = getConfig();
        if (player.isOp() && cfg.deopOnJoin) {
            event.setCancelled(true);
            player.sendMessage(ChatColor.RED + "Please wait until you've fully joined.");
        }
    }
```

me4502: Coming back to this, that workaround doesn't really fix the issue. It just breaks the concept of op status entirely.

Currently this alternate event appears to not be in a state where we can move to it, as it breaks a widely used safety feature with no safe or equivalent workaround

PR #2241, opened 2025-11-09
Localize user-facing WorldGuard messages via messages_en.yml

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
Extracted hard-coded user-facing strings from core WorldGuard classes into messages_en.yml. This lets every server owner or translator adjust the plugin’s output without touching Java code, making custom translations straightforward.

image


me4502:
Was this PR done with AI? This touches a substantial amount of messages that should _not_ be localised, such as hardcoded constants.

Any setup like this should be integrated into the existing translation system, that supports automatic translations rather than requiring each server owner to do it themselves. The reason it's not currently setup everywhere is the current command system doesn't support translations, and no other command system that does has the flexibility that WG needs.

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
I focused on moving only the strings that are safe to translate, like log messages, warnings, and report text. I left command annotations and any constants alone so the current command system stays intact. WorldGuard has been around for more than ten years, and it feels important that players can finally see messages in their own language without risking stability. I’m running tests right now to make sure everything works as expected, and I’ll reach out immediately if anything unexpected shows up.

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
The translation[ into Russian](https://github.com/Black-Minecraft-com/WorldGuard/blob/version/7.0.x/worldguard-bukkit/src/main/resources/defaults/lang/messages_ru.yml) was done with the help of AI.

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
@me4502 we’ve run the necessary tests, and all the phrases I moved into the translation files work perfectly.
Could you please merge the project?

b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0b0:
image
image

image

me4502:
As I said in my previous comment- any changes like this should be integrated into the existing translation system (TranslationManager class), instead of adding another one on top. I don't recall to what degree it's currently setup in WorldGuard, but it'd just look like this generally (https://github.com/EngineHub/CraftBook/blob/five/craftbook-core/src/main/java/org/enginehub/craftbook/CraftBook.java#L68).

Also as a change like this alters API, it should be targeting the `master` branch, not `7.0.x`

Joo200:
I don't agree on making log lines translations. We already have some issues when we support users with translated user output. With translated it would be a big PITA to provide good support.

PR #2247, opened 2025-12-19
Fixed disappearing items

daniel-skopek:
When the item-drop region flag is set to deny and the inventory is full, the armor item disappears when dropped.

This pull request fixes that.


GG-MD (commented):
`player.sendMessage("1");` remove.

wizjany:
this kind of thing needs to be done in the server itself as part of event cancellation handling. messing with state like this, especially in ways that could directly counteract what WG is trying to do, doesn't belong here.

PR #2249, opened 2025-12-27
Updated material definitions and dependencies for 1.21.11

aerulion:
- Updated Mojang-provided libraries to the correct version
- Added new Material and SpawnEgg definitions (the larger diff in the spawn egg method is just due to fixing the alphabetical order which got messed up at some point)
- Fixed some missing item materials i missed during the previous commit, since I did not know items should also be covered

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-worldedit = "7.3.9"
+worldedit = "7.3.18"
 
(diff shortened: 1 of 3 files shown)


me4502 (approved):
Thank you for this :)

me4502:
I'm doing a semi weird thing, partially merging this including only 1.21.10 compatible changes- making a release, then re-merging the rest of it in

PR #2259, opened 2026-03-05
Basic Folia Support

Euphillya:
Following the [basic Folia support for WorldEdit](https://github.com/EngineHub/WorldEdit/pull/2905), this PR adds basic Folia support.
Some testing still needs to be done.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-worldedit = "7.3.18"
+worldedit = "7.4.1-SNAPSHOT"
 
 # Minimum versions we apply to make dependencies support newer Java
 minimumAsm = "9.7"
@@ -31,7 +31,7 @@ paperApi = "io.papermc.paper:paper-api:1.21.11-R0.1-SNAPSHOT"
 paperLib = "io.papermc:paperlib:1.0.8"
 
 dummypermscompat = "com.sk89q:dummypermscompat:1.10"
-bstats-bukkit = "org.bstats:bstats-bukkit:2.2.1"
+bstats-bukkit = "org.bstats:bstats-bukkit:3.2.1"
 
 prtree = "org.khelekore:prtree:1.5.0"
 
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-worldedit = "7.4.1-SNAPSHOT"
+worldedit = "7.3.18"
 
 # Minimum versions we apply to make dependencies support newer Java
 minimumAsm = "9.7"
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
@@ -119,14 +119,27 @@ public class PlayerMoveListener extends AbstractListener {
                     current.eject();
                     vehicle.setVelocity(new Vector());
                     if (vehicle instanceof LivingEntity) {
-                        PaperLib.teleportAsync(vehicle, override.clone());
+                        if (getPlugin().isFolia()) {
+                            PaperLib.teleportAsync(vehicle, override.clone());
+                        } else {
+                            vehicle.teleport(override.clone());
+                        }
                     } else {
-                        PaperLib.teleportAsync(vehicle, override.clone().add(0, 1, 0));
+                       if (getPlugin().isFolia()) {
+                           PaperLib.teleportAsync(vehicle, override.clone().add(0, 1, 0));
+                       } else {
+                           vehicle.teleport(override.clone());
+                       }
                     }
                     current = current.getVehicle();
                 }
 
-                PaperLib.teleportAsync(player, override.clone().add(0, 1, 0));
+                if (getPlugin().isFolia()) {
+                    PaperLib.teleportAsync(player, override.clone().add(0, 1, 0));
+                } else {
+                    player.teleport(override.clone().add(0, 1, 0));
+                }
+
 
                 if (getPlugin().isFolia()) {
                     player.getScheduler().runDelayed(getPlugin(), scheduledTask -> PaperLib.teleportAsync(player, override.clone().add(0, 1, 0)), null, 1);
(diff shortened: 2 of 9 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/util/report/SchedulerReport.java
```diff
@@ -51,6 +52,7 @@ public Optional<Field> load(Class<?> clazz) throws Exception {
 
     public SchedulerReport() {
         super("Scheduler");
+        if (WorldGuardPlugin.inst().isFolia()) return; // Todo Add support Schedulers
```
me4502: Please use `{ }` around this return statement

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
```diff
@@ -118,16 +119,33 @@ public void onPlayerMove(PlayerMoveEvent event) {
...
-                        vehicle.teleport(override.clone());
+                        if (getPlugin().isFolia()) {
+                            PaperLib.teleportAsync(vehicle, override.clone());
+                        } else {
+                            vehicle.teleport(override.clone());
+                        }
                     } else {
-                        vehicle.teleport(override.clone().add(0, 1, 0));
+                       if (getPlugin().isFolia()) {
+                           PaperLib.teleportAsync(vehicle, override.clone().add(0, 1, 0));
```
me4502: For cases like this where there's actual calculation done in here, and the other one below, IMO it makes sense to extract this location variable to before the folia check so that the logic isn't duplicated. That way if it were to ever change in the future, there's less chance of it going out of sync.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
```diff
@@ -118,16 +119,33 @@ public void onPlayerMove(PlayerMoveEvent event) {
...
+                       } else {
+                           vehicle.teleport(override.clone().add(0, 1, 0));
+                       }
                     }
                     current = current.getVehicle();
                 }
 
-                player.teleport(override.clone().add(0, 1, 0));
+                if (getPlugin().isFolia()) {
+                    PaperLib.teleportAsync(player, override.clone().add(0, 1, 0));
```
me4502: (This being the other point referenced in the last comment)

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardVehicleListener.java
```diff
@@ -66,10 +67,18 @@ public void onVehicleMove(VehicleMoveEvent event) {
...
+                        if (getPlugin().isFolia()) {
+                            PaperLib.teleportAsync(vehicle, event.getFrom());
+                        } else {
+                            vehicle.teleport(event.getFrom());
+                        }
                         if (Locations.isDifferentBlock(lastValid, BukkitAdapter.adapt(event.getFrom()))) {
                             Vector dir = player.getLocation().getDirection();
-                            player.teleport(BukkitAdapter.adapt(lastValid).setDirection(dir));
+                            if (getPlugin().isFolia()) {
+                                PaperLib.teleportAsync(player, BukkitAdapter.adapt(lastValid).setDirection(dir));
```
me4502: This is also likely a good candidate to extract the location variable

me4502 (changes requested):
Thanks for this. It appears to mostly follow along with the same goals from the WorldEdit Folia support.

I've left a few comments on some parts that I feel would be worth cleaning up

Euphillya:
That should be good, I also took the liberty of removing duplicates for `event.getFrom()` and `event.getTo()` in the WorldGuardVehicleListener.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
```diff
@@ -118,16 +119,37 @@ public void onPlayerMove(PlayerMoveEvent event) {
                     current.eject();
                     vehicle.setVelocity(new Vector());
                     if (vehicle instanceof LivingEntity) {
-                        vehicle.teleport(override.clone());
+                        Location vehicleTeleportLocation = override.clone();
+                        if (getPlugin().isFolia()) {
+                            PaperLib.teleportAsync(vehicle, vehicleTeleportLocation);
```
Joo200, line 124: I don't think there is a reason to use PaperLib only for folia. Imo we could always use PaperLib which provides fallback stuff for spigot.

Euphillya: The potential problem would be that there could be a desynchronization between several teleportations on the Paper or Bukkit servers, which are currently a synchronous call
An issue has been created for WorldEdit regarding an API; if we know it won't cause any problems, I can merge using PaperLib for Bukkit like Folia.

me4502: I feel for now, it's better to just do minimal changes for Folia support. We can do better testing as to whatever async teleports work well on Paper later, but I feel that part is out of scope for this PR

gradle/libs.versions.toml
```diff
@@ -1,5 +1,5 @@
 [versions]
-worldedit = "7.3.18"
+worldedit = "7.4.1-SNAPSHOT"
```
me4502: Is this actually using anything from the new WorldEdit release? Generally we try to not target snapshots (if it's using something from 7.4, building against `7.4.0` would be preferred)

me4502 (approved):
Thanks for this, overall IMO these changes are a good basis for Folia support in WorldGuard.

gradle/libs.versions.toml
Euphillya: I simply forgot to change it back to how it was before, because I originally wanted to use WorldEditPlugin#isFolia but it's protected.

me4502: Would you be able to please revert that change then? Also possibly the bstats one below, unless that version bump is related to Folia support in some way

Euphillya: It's been reverted to the previous state. 
And for Bstats, the version upgrade includes Folia support.

PR #2265, opened 2026-03-23
Only store originalGameMode if NULL

### LINKED ISSUE #2264: Gamemode flags resolve incorrectly on teleport when moving out of nested regions
### WorldEdit Version

7.3.18

### WorldGuard Version

7.0.15

### Platform Version

purpur-1.21.11-2566.jar

### Confirmations

### Bug Description

If a player is inside a higher-priority region (`gamemode` set to `survival`) nested within a lower-priority region (`gamemode` set to `creative`) and teleports directly outside both regions, WorldGuard applies the outer region’s gamemode (creative) instead of reverting to the world’s default.

This only occurs when teleporting; walking between regions resolves gamemodes correctly.

### Expected Behavior

If the destination is not within any region with a gamemode flag, the player should revert to the world’s default gamemode.

### Reproduction Steps

1. Set the world’s default gamemode to Survival
2. Create a region `inner` with:
    * `gamemode` = `survival`
    * `priority` = `2`
3. Create a region `outer` that fully contains `inner`, with:
    * `gamemode` = `creative`
    * `priority` = `1`
4. Stand inside inner
5. Teleport to a location outside both regions
6. You'll now incorrectly be in Creative mode

_No response_

### Anything Else?

Note that this bug can also be reproduced on the latest beta release of WorldGuard.
This report is from a Purpur server, but also can be reproduced on the latest Paper and Spigot.

BriteBytes:
Fixes #2264

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/session/handler/GameModeFlag.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/session/handler/GameModeFlag.java
@@ -59,7 +59,9 @@ public class GameModeFlag extends FlagValueChangeHandler<GameMode> {
     private void updateGameMode(LocalPlayer player, @Nullable GameMode newValue, World world) {
         if (!getSession().getManager().hasBypass(player, world) && newValue != null) {
             if (player.getGameMode() != newValue) {
-                originalGameMode = player.getGameMode();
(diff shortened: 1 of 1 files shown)


me4502:
Thanks for this PR, it works from my testing.

PR #2271, opened 2026-05-01
Updated alias command names

Tavies:
Using WorldGuard specific aliases such as wglocate etc allows users to avoids having conflicts with native minecraft commands such as locate.


me4502:
This is a breaking change so isn't really something we can merge sorry. While I am not personally a fan of these commands, they do have substantial use so we can't just rename them.

A PR that adds a simple config option to disable them (that defaults to them being enabled) would be accepted, however nothing that breaks expected functionality by default can really be merged.

PR #2275, opened 2026-05-14
Add lava-harden flag

TomyLobo:
This covers the forming of cobblestone, stone, obsidian and basalt.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
@@ -478,7 +478,8 @@ public class WorldGuardBlockListener extends AbstractListener {
 
         WorldConfiguration wcfg = getWorldConfig(event.getBlock().getWorld());
 
-        Material type = event.getNewState().getType();
+        Material oldType = event.getBlock().getType();
+        Material newType = event.getNewState().getType();
 
         if (event instanceof EntityBlockFormEvent) {
             if (((EntityBlockFormEvent) event).getEntity() instanceof Snowman) {
@@ -490,7 +491,7 @@ public class WorldGuardBlockListener extends AbstractListener {
             return;
         }
 
-        if (type == Material.ICE) {
+        if (newType == Material.ICE) {
             if (wcfg.disableIceFormation) {
                 event.setCancelled(true);
                 return;
@@ -502,7 +503,7 @@ public class WorldGuardBlockListener extends AbstractListener {
             }
         }
 
-        if (type == Material.SNOW) {
+        if (newType == Material.SNOW) {
             if (wcfg.disableSnowFormation) {
                 event.setCancelled(true);
                 return;
@@ -522,7 +523,7 @@ public class WorldGuardBlockListener extends AbstractListener {
             }
         }
 
-        if (Materials.isUnwaxedCopper(event.getBlock().getType())) {
+        if (Materials.isUnwaxedCopper(oldType)) {
             if (wcfg.disableCopperBlockFade) {
                 event.setCancelled(true);
                 return;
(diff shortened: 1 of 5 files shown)


worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
```diff
@@ -533,6 +535,19 @@ public void onBlockForm(BlockFormEvent event) {
                 return;
             }
         }
+
+        if (event.getBlock().getType() == Material.LAVA && !newState.getBlock().isLiquid()) {
```
me4502: Does this need to also check the flowing lava block, or does this cover all cases?

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
```diff
@@ -478,7 +479,8 @@ public void onBlockForm(BlockFormEvent event) {
 
         WorldConfiguration wcfg = getWorldConfig(event.getBlock().getWorld());
 
-        Material type = event.getNewState().getType();
+        BlockState newState = event.getNewState();
```
wizjany: is this var (or the import) actually used anywhere? or is it a leftover diff from a previous commit

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
```diff
@@ -533,6 +535,19 @@ public void onBlockForm(BlockFormEvent event) {
                 return;
             }
         }
+
+        if (event.getBlock().getType() == Material.LAVA && !Materials.isLiquid(newState.getType())) {
```
me4502: newState.getType() here should probably be replaced with the `type` variable that's arleady extracted – repeated type lookups have a surprisingly large performance overhead

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
TomyLobo: yeah that isn't needed anymore. fixed it.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
TomyLobo: yeah that's an oversight, fixed

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
TomyLobo: this covers flowing lava

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
```diff
@@ -533,6 +533,19 @@ public void onBlockForm(BlockFormEvent event) {
                 return;
             }
         }
+
+        if (event.getBlock().getType() == Material.LAVA && !Materials.isLiquid(type)) {
```
me4502: As you are adding a second thing that grabs `event.getBlock().getType()`, it's likely worth extracting this to a variable and replacing the other usages of it. This event can be fired a bunch, and doubling the getType call counts can have a surprisingly large impact on performance

me4502 (changes requested):

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/WorldGuardBlockListener.java
TomyLobo: done

me4502 (approved):

PR #2281, opened 2026-05-29
Fix SessionManager world retention after player quit and world unload

ikoliHU:
## Problem

`SessionManager` can keep unloaded worlds reachable through cached player/session state.

This is visible with island/Slime worlds. After a player leaves and the world is unloaded, a heap dump can still show the world retained through:

```
SlimeInMemoryWorld
-> SlimeLevelInstance
-> ServerPlayer
-> CraftPlayer
-> WorldPlayer (LocalPlayer)
-> WorldPlayerTuple
-> com.google.common.cache.LocalCache$StrongAccessWriteEntry
-> com.sk89q.worldguard.bukkit.session.SessionManager
```

There are two relevant cache paths:

1. The bypass cache uses `LoadingCache<WorldPlayerTuple, Boolean>`. `WorldPlayerTuple` strongly references both the `World` and the `LocalPlayer`, which also keeps the platform player object reachable. The cache uses `expireAfterWrite(2s)`, but Guava removes expired entries lazily during cache operations. If the cache is idle, an expired entry can remain reachable longer than intended.

2. The session cache uses `Cache<CacheKey, Session>`. The key weakly references the player, but the `Session` value can still retain the last world through `Session#lastValid`. On quit the session is uninitialized, but the cache entry is not removed.

## Changes

* Add `SessionManager#forget(LocalPlayer)`.
* Add `SessionManager#forgetWorld(World)`.
* Invalidate matching session and bypass-cache entries from those methods.
* Add cache cleanup in `AbstractSessionManager`.
* Call `forget(player)` after session uninitialization on `PlayerQuitEvent`.
* Call `forgetWorld(world)` on `WorldUnloadEvent`.
* Run cache cleanup on each session tick.

## Result

* Player quit removes the related session and bypass-cache entries.
* World unload removes bypass-cache entries for that world.
* Expired bypass-cache entries are cleaned up on the next session tick instead of waiting for another unrelated cache operation.
* Plugins using short-lived worlds can call `forget` or `forgetWorld` directly instead of accessing WorldGuard internals.

## Testing

* `:worldguard-core:compileJava` passed.
* `:worldguard-bukkit:compileJava` passed.
* `:worldguard-core:checkstyleMain` passed.
* `:worldguard-bukkit:checkstyleMain` passed.
* Repeated player enter, leave/disconnect, and world unload cycles no longer leave `SlimeInMemoryWorld` or `SlimeLevelInstance` retained through `SessionManager` in heap dumps.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
@@ -169,6 +169,11 @@ public class PlayerMoveListener extends AbstractListener {
         }
 
         session.uninitialize(localPlayer);
+
+        // Drop the cached session and bypass entries so the manager does not
+        // keep the (now offline) player and their last world reachable until
+        // the caches expire.
+        WorldGuard.getInstance().getPlatform().getSessionManager().forget(localPlayer);
     }
 
     /**
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/PlayerMoveListener.java
@@ -169,11 +169,6 @@ public class PlayerMoveListener extends AbstractListener {
         }
 
         session.uninitialize(localPlayer);
-
-        // Drop the cached session and bypass entries so the manager does not
-        // keep the (now offline) player and their last world reachable until
-        // the caches expire.
-        WorldGuard.getInstance().getPlatform().getSessionManager().forget(localPlayer);
     }
 
     /**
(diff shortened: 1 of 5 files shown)


content:
Awesome pull request!

me4502:
Thanks for the PR, I haven't had a chance to look through it properly, but from a quick look it seems to break some functionality. Sessions are meant to stay around for a while after the player leaves, so clearing them from the cache breaks some functionality there. Clearing the world itself is likely fine, but not removing the session itself. That should likely instead be handled by setting lastValid to null in the uninitialize function, rather than clearing the player from the cache.

ikoliHU:
Thanks for the review, you were right. I pushed a follow-up with the suggested approach.

Updated from the previous version:

* Removed the `forget(localPlayer)` call on player quit, so sessions stay cached as intended.
* Removed the now-unused `SessionManager#forget(LocalPlayer)`.
* `Session#uninitialize` now clears `lastValid` and `lastRegionSet`, so the lingering session no longer holds world/region references.

I kept the world-level cleanup:

* `forgetWorld(World)` + `WorldUnloadEvent` listener for the bypass cache.
* Periodic cache cleanup for idle expired bypass-cache entries.

I tested these changes as well, and my tests still passed. `SlimeInMemoryWorld` instances no longer got stuck.

I hope this is the change you had in mind.

PR #2285, opened 2026-06-24
perf: avoid unnecessary isNPC checks

RVSkeLe:
Check whether an entity is a Player before calling `isNPC()` in player-only listeners, avoiding unnecessary NPC checks.

No behavioral changes intended.


me4502 (approved):
Thanks for this, I do agree a simple instanceof check is likely faster than the metadata lookup of isNPC. All of the places you've changed appear to be safe.

PR #2295, opened 2026-07-27
Add Cause tracking for SulfurCube

### LINKED ISSUE #2294: Interaction with sulfur cubes are not covered by WorldGuard
### WorldEdit Version

7.4.4+7546-f9e033f

### WorldGuard Version

7.0.17+2370-e42d8bc

### Platform Version

Paper 26.2-63-3a55f62 (MC: 26.2)

### Confirmations

### Bug Description

Worldguard doesn't prevent any interaction with sulfur cubes in protected regions for non-owner/members. You can hit cubes, place, extract and ignite blocks inside sulfur cube. Also WG doesn't cancel block damage from sulfur cube's tnt.

Explosion of cube's dymanite can destroy nearby blocks and trigger another tnt. Normal tnt, even triggered by sulfur cube's tnt, doesn't break any blocks.

### Expected Behavior

Strangers (not owners and not members of protected region) can't interact with sulfur cube. Explosion of sulfur cube should not damage region's blocks.

### Reproduction Steps

1. Enter someone's region.
2. Check that you can't place blocks inside region and "normal" tnt dropped from above doesn't harm region.
3. Go outside and spawn sulfur cube with `Bucket of Sulfur Cube`.
4. Push sulfur cube into protected region.
5. Place dirt into cube.
6. Push sulfur cube by hitting it.
7. Extract dirt from cube with `Shears`.
8. Place tnt into cube.
9. Ignine tnt with `Flint and Steel`.
10. Wait for explosion.

https://paste.enginehub.org/iRCju7Lmq.report

### Anything Else?

Here is screencast: https://www.youtube.com/watch?v=XehSzUPOzgQ

Firemoon777:
This PR fixes #2294.

Explosive archetype of sulfur cube is dangerous by default and cause damage to blocks. Sulfur cubes can transform to explosive by consuming a TNT block: from player's direct interaction or by grabbing dropped item. Damage from sulfur cube may be disabled with `other-explosion: DENY` flag, but there is no fun to disable it everywhere.

This PR brings more consistent approach: sulfur cube with TNT acts like normal tnt and explosive minecarts at once. I added Cause tracking for every interaction. Insertion block by player will track the root cause of possible destruction. This means that sulfur cube with TNT inserted by non-MEMBER and pushed to protected region would not harm any blocks. Even feeding a cube with TNT-item within protected region and pushing to any redstone block will be safe for region, because cube itself will treated as non-MEMBER. But cube, created by MEMBER or OWNER still can explode blocks if flag `tnt` is not set to DENY.

But there are still more caveates in latest version of Paper:
- WorldGuard prevents stealing cube with bucket, but sulfur cube will disappear. It's only client-side, bug reported: https://github.com/PaperMC/Paper/issues/14113
- Player still can spawn cube within protected region if player aims into another entity and pointing "through" entity on near block. According to logs every interaction event is cancelled, but cube still spawns. Not sure if it a WorldGuard issue.
- Cube's explosion can trigger nearby explosive sulfur cubes. Damage events are cancelled by WorldGuard as well, but cubes somehow don't care. Not sure if it a WorldGuard issue again

Also not sure if this PR meets all requirements/code style, but I tried to do my best.

### DIFF
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/util/Entities.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/util/Entities.java
@@ -43,6 +43,7 @@ import org.bukkit.entity.Player;
 import org.bukkit.entity.Projectile;
 import org.bukkit.entity.SpectralArrow;
 import org.bukkit.entity.Steerable;
+import org.bukkit.entity.SulfurCube;
 import org.bukkit.entity.TNTPrimed;
 import org.bukkit.entity.Tameable;
 import org.bukkit.entity.Vehicle;
@@ -78,7 +79,9 @@ public final class Entities {
      * @return true if TNT based
      */
     public static boolean isTNTBased(Entity entity) {
-        return entity instanceof TNTPrimed || entity instanceof ExplosiveMinecart;
+        return entity instanceof TNTPrimed
+                || entity instanceof ExplosiveMinecart
+                || entity instanceof SulfurCube sulfurCube && sulfurCube.canExplode();
     }
 
     /**
@@ -265,6 +268,7 @@ public final class Entities {
             case Wither wither -> Flags.WITHER_DAMAGE;
             case Creeper creeper -> Flags.CREEPER_EXPLOSION;
             case TNTPrimed tnt -> Flags.TNT;
+            case SulfurCube sulfurCube -> Flags.TNT;
             case ExplosiveMinecart minecart -> Flags.TNT;
             case EnderDragon dragon -> Flags.ENDERDRAGON_BLOCK_DAMAGE;
             case null, default -> Flags.OTHER_EXPLOSION;
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/cause/Cause.java
@@ -40,6 +40,7 @@ import org.bukkit.entity.Firework;
 import org.bukkit.entity.LightningStrike;
 import org.bukkit.entity.Player;
 import org.bukkit.entity.Projectile;
+import org.bukkit.entity.SulfurCube;
 import org.bukkit.entity.TNTPrimed;
 import org.bukkit.entity.Tameable;
 import org.bukkit.entity.Vehicle;
@@ -117,7 +118,7 @@ public final class Cause {
             return false;
         }
 
-        if (object instanceof TNTPrimed || object instanceof Vehicle) {
+        if (object instanceof TNTPrimed || object instanceof Vehicle || object instanceof SulfurCube) {
             if (!PaperLib.isPaper()) {
                 return false;
             }
--- a/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
+++ b/worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/listener/EventAbstractionListener.java
@@ -83,6 +83,7 @@ import org.bukkit.entity.LivingEntity;
 import org.bukkit.entity.Mob;
 import org.bukkit.entity.Painting;
 import org.bukkit.entity.Player;
+import org.bukkit.entity.SulfurCube;
 import org.bukkit.entity.Tameable;
 import org.bukkit.entity.ThrownPotion;
 import org.bukkit.entity.WindCharge;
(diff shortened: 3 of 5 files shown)


me4502:
Thanks for this PR :) – Currently I'm wanting to hold off on bumping our requirement to 26.2 mostly to have one more final 26.1-compatible release, so I'll wait for that point before doing a proper review.

From a very quick look though there are two main questions:

* Is this adding usage of any Paper-specific APIs? (eg, will this break Spigot compatibility)
* Are there any non-TNT ways to make a sulfur cube explode? Just noticed a lot of the checks are basically just treating an explosion as TNT if a sulfur cube explodes, so just double checking that it's a safe/correct assumption

Firemoon777:
>Is this adding usage of any Paper-specific APIs? (eg, will this break Spigot compatibility)

No, I guess no Paper-specific APIs were used. `SulfurCube` Entity exists in Spigot and I used none of its inherited interfaces, just checking with `instanceof`. 

>Are there any non-TNT ways to make a sulfur cube explode?

As for now, and according to [Minecraft Wiki](https://minecraft.wiki/w/Sulfur_Cube#Special_archetypes), explosive archetype is achieved by consuming items with tag `#minecraft:sulfur_cube_archetype/explosive`. Again, according to [Minecraft Wiki Item Tag page](https://minecraft.wiki/w/Item_tag_(Java_Edition)#sulfur_cube_archetype/explosive) there is only one item with this tag — `minecraft:tnt`. 

Moreover, sulfur cube can't explode by its own, like a creeper. Cube loses ability to move after consuming item, so, cube requires any external action like igniting or pushing to redstone blocks to perform an explosion. That's why I treating sulfur cube as TNT in explosion-related events

me4502:
Makes sense, thanks :) – Are you able to please now rebase? I've made the required changes for WorldGuard to target MC 26.2, which will likely conflict with some parts of this PR as it was also partially done here.

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/util/Entities.java
```diff
@@ -265,9 +268,20 @@ public static boolean isPluginSpawning(CreatureSpawnEvent.SpawnReason spawnReaso
...
     }
+
+    /**
+     * Test whether an entity is Sulfur Cube.
+     *
+     * @param entity the entity
+     * @return true if sulfur cube
+     */
+    public static boolean isSulfurCube(Entity entity) {
+        return entity instanceof SulfurCube;
```
me4502: Do we actually need this method here? It seems to just be used once, and is just a simple instanceof check that's unlikely to ever expand in the future. If this was "is an explodeable sulfur cube" it'd make more sense to have a utility method like this, but I'm unsure if it has much merit existing in its current state

me4502 (changes requested):

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/util/Entities.java
Firemoon777: Not really. I've converted it to `isExplodableSulfurCube(Entity)` and used it in `isTNTBased(Enitity)` (which marked as unused in my IDE)

But if it still useless in you opition, I can remove it completly, of course

Firemoon777:
Sure, I rebased my branch from current upstream `version/7.0.x`

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/util/Entities.java
me4502: Thanks, if it's just the once case still it's probably worth dropping entirely tbh (while replacing the conditional with `entity instanceof SulfurCube sulfurCube && sulfurCube.canExplode();` to make it cleaner)

Firemoon777: Yes, this is the once case of using this method.
I just removed it

worldguard-bukkit/src/main/java/com/sk89q/worldguard/bukkit/util/Entities.java
```diff
@@ -78,7 +79,9 @@ public static boolean isTamed(@Nullable Entity entity) {
      * @return true if TNT based
      */
     public static boolean isTNTBased(Entity entity) {
-        return entity instanceof TNTPrimed || entity instanceof ExplosiveMinecart;
+        return entity instanceof TNTPrimed
+                || entity instanceof ExplosiveMinecart
+                || (entity instanceof SulfurCube && ((SulfurCube)entity).canExplode());
```
me4502: Are you able to please apply the style change from my last comment here?

`entity instanceof SulfurCube sulfurCube && sulfurCube.canExplode();`

Or alternatively just add a space between the (SulfurCube) cast and the entity, as the current one will produce checkstyle warnings

Firemoon777: Oh, my bad. Fixed with your suggestion

me4502 (approved):
Thank you! :)

PR #2298, opened 2026-08-03
Avoid unnecessary synchronization in QueryCache#queryContains

1wairesd:
Use a lock-free ConcurrentHashMap.get() before falling back to compute() to avoid bucket-level locking when the result is already cached. On the hot path (repeated queries to the same block position within a tick), this eliminates contention on the CHM segment lock entirely, reducing overhead to a single volatile read.

### DIFF
--- a/worldguard-core/src/main/java/com/sk89q/worldguard/protection/regions/QueryCache.java
+++ b/worldguard-core/src/main/java/com/sk89q/worldguard/protection/regions/QueryCache.java
@@ -57,6 +57,16 @@ public class QueryCache {
         checkNotNull(option);
 
         CacheKey key = new CacheKey(location);
+
+        // Fast path: avoid locking if the result is already cached
+        Map<QueryOption, ApplicableRegionSet> existing = cache.get(key);
+        if (existing != null) {
+            ApplicableRegionSet result = existing.get(option);
+            if (result != null) {
+                return result;
+            }
+        }
+
         return cache.compute(key, (k, v) -> option.createCache(manager, location, v)).get(option);
     }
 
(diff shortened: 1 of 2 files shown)


SirYwell:
It looks like the lock ensures that access to the inner map is synchronized, which seems to be required.

1wairesd:
Fixed the thread-safety concern by replacing EnumMap with ConcurrentHashMap for the inner option map in all three QueryOption.createCache() implementations. The computeIfAbsent calls on the inner map are now safe, and the lock-free get() fast-path in QueryCache.queryContains() is correct.

worldguard-core/src/main/java/com/sk89q/worldguard/protection/regions/QueryCache.java
```diff
@@ -57,6 +57,16 @@ public ApplicableRegionSet queryContains(RegionManager manager, Location locatio
...
         CacheKey key = new CacheKey(location);
+
+        // Fast path: avoid locking if the result is already cached
+        Map<QueryOption, ApplicableRegionSet> existing = cache.get(key);
+        if (existing != null) {
+            ApplicableRegionSet result = existing.get(option);
+            if (result != null) {
+                return result;
+            }
+        }
```
octylFractal, line 68: Instead of this, can we split the `createCache` method into two parts, one that returns the appropriate map, and one that fills in the information with `computeIfAbsent` or whatever is typically used in the `!= null` branch? That way, instead of still having lock contention for every initial query, we get out of the `computeIfAbsent` quickly and do a follow-up one only on the smaller map if necessary.

It might not be perfectly that clean, but I think we should avoid having expensive logic inside a `compute` or this fix won't help that much.

PR #2307, opened 2026-09-18
Avoid NPE when clearing a region flag

### LINKED ISSUE #2305: NullPointerException when setting flag to "none" in RegionCommands.setFlag
### WorldEdit Version

7.4.5+7590-b8dc4c1

### WorldGuard Version

7.0.18+2392-fa605e6

### Platform Version

Purpur 26.2-2627-HEAD@f284e3d

### Confirmations

### Bug Description

When setting a WorldGuard flag to `none` to remove it, the plugin successfully clears the flag and saves the data in the background, but throws a `NullPointerException` in the server console/chat. This happens because the command framework attempts to invoke `.toString()` on the `null` value returned by `RegionCommands.setFlag(...)` to build the console/chat feedback.

Copy from the log: 
<html>
<body>

[20:25:40] [Server thread/INFO]: Cannot invoke "Object.toString()" because the return value of "com.sk89q.worldguard.commands.region.RegionCommands.setFlag(com.sk89q.worldguard.protection.regions.ProtectedRegion, com.sk89q.worldguard.protection.flags.Flag, com.sk89q.worldedit.extension.platform.Actor, String)" is null
--
  | [20:25:44] [WorldGuard Region I/O/INFO]: [WorldGuard] Region data changes made in 'world' have been background saved

</body>
</html>

### Expected Behavior

The plugin should successfully clear the flag, gracefully handle the `null` return value from the cleared state, and send a standard confirmation message without throwing any errors in the console.

### Reproduction Steps

1. Create a region and assign a flag to a specific value (e.g., `/rg flag testregion exit deny`).
2. Attempt to remove the flag by setting it to none (e.g., `/rg flag testregion exit none`).
3. Check the server console to observe the `Cannot invoke "Object.toString()" ... is null` error, followed by the background save message.

_No response_

### Anything Else?

_No response_

HP-network:
When a state flag is set to `none`, parsing intentionally returns null after clearing the flag. The command feedback path called `toString()` on that value, so the command completed but logged an NPE.

Keep the existing normalized output for non-null values and report `none` for the clear operation.

Fixes #2305

Tested with `./gradlew :worldguard-core:test`.


worldguard-core/src/main/java/com/sk89q/worldguard/commands/region/RegionCommands.java
```diff
@@ -597,7 +597,8 @@ public void flag(CommandContext args, Actor sender) throws CommandException {
         if (value != null) {
             // Set the flag if [value] was given even if [-g group] was given as well
             try {
-                value = setFlag(existing, foundFlag, sender, value).toString();
+                Object parsedValue = setFlag(existing, foundFlag, sender, value);
+                value = parsedValue == null ? "none" : parsedValue.toString();
```
me4502: This change is treating it like the value "none" is being set on the region, whereas the flag is being removed. This should ideally trigger a message such as the one on like 621 (https://github.com/EngineHub/WorldGuard/pull/2307/changes#diff-cd6887d878ad3f9bebfe64edf07696bb25bb4f33a91877bf814acefd800243caL621). So maybe rather than just subbing in the word "none", setting `value` to null so that it can be better handled with proper messaging later on.

me4502 (changes requested):
Thanks for this PR! :)

The original issue misconstrued the problem a bit, this isn't the recommended nor documented way to unset flags on a region, but worked due to some legacy parsing code that happened to line up with command handling. As far as I can tell, some LLMs hallucinated `none` as the correct keyword to unset a flag, which led to people starting to use the wrong command.

IMO the goal here should be to prevent it from actually erroring when using this alternate command, rather than changing any behaviour beyond that point. Currently this PR uses different messaging when setting a flag to the `none` value compared to unsetting a flag, and ideally the `none` value should be treated the same as the intended command to unset a flag. I've left a comment on a way that I feel would fix that side of things :)

HP-network:
Thanks, that distinction makes sense. `none` is taking the clear/remove path, so it should keep the existing unset feedback rather than be treated as a value to set. I'll adjust the command handling accordingly and rerun the core tests.

